PR #1, opened 2020-08-06
RBACUsers support

tjrivera:
Copied over from https://github.com/hbagdi/go-kong/pull/29

I added TODOs here around integration with other Kong endpoints that I anticipate including in subsequent patches. (Roles, Permissions, etc.)


kong/kong.go
```diff
@@ -43,6 +43,7 @@ type Client struct {
 	Targets        *TargetService
 	Workspaces     *WorkspaceService
 	Admins         *AdminService
+	Users          *RBACUserService
```
hbagdi, line 46: s/Users/RBACUsers

PR #2, opened 2020-08-06
Workspace Entities Support

tjrivera:
Moved from https://github.com/hbagdi/go-kong/pull/30

This patch provides workspace entities support. The use case here is that in some cases, we would like to migrate services from one workspace to another, and to do it without the possibility of downtime. Additionally, we've found that the only way to remove an Admin from a workspace when they are added with the oidc plug-in enabled, is to manipulate the workspace entities associated with that Admin. Deleting them from a workspace ends up deleting them from Kong entirely. I understand this behavior may change in Kong 2.x


hbagdi:
> This patch provides workspace entities support. The use case here is that in some cases, we would like to migrate services from one workspace to another, and to do it without the possibility of downtime. Additionally, we've found that the only way to remove an Admin from a workspace when they are added with the oidc plug-in enabled, is to manipulate the workspace entities associated with that Admin. Deleting them from a workspace ends up deleting them from Kong entirely. I understand this behavior may change in Kong 2.x

That is unfortunate and I want to resist adding this change in because of future changes to this API in Kong Enterprise but at the same time I do understand the need here, so let's move ahead.

Did you consider not introducing a new service and rather adding additional methods in WorkspaceService?

tjrivera:
Agreed, we'd rather not have to manage entities if we can as we recognize the brittle nature of making those types of updates. I split this off as another service based on [your feedback on an earlier PR](https://github.com/hbagdi/go-kong/pull/18#issuecomment-629458730) but recognize that it probably makes sense for these methods to be on the workspace service and not necessarily an entirely new service. I'm happy to refactor things to move entity operations to the workspace service if you agree that that makes sense.

hbagdi:
> split this off as another service based on your feedback on an earlier PR but recognize that it probably makes sense for these methods to be on the workspace service and not necessarily an entirely new service. I'm happy to refactor things to move entity operations to the workspace service if you agree that that makes sense.

Apologies for the miscommunication. In that comment, I wanted to communicate that we handle one area at a time and keep the diffs small to sanely review them and move faster with fewer bugs.
I think it makes sense to keep the entities funcs on the WorkspaceService itself because they relate to Workspaces more than a standalone entity.

tjrivera:
@hbagdi, I moved the entities work into the workspace service. this should be ready for review. I plan on submitting the endpoint/entity permissions work soon, followed by roles.

PR #8, opened 2020-10-19
feat: rich HTTP errors

mflendrich:
This PR:
- improves error messages returned by HTTP calls to Kong Admin API by including the `message` from the response body,
- ensures that all users test errors for being equal to 404 by means of the public API.

Inspired by https://github.com/Kong/deck/issues/218#issuecomment-712192401.

### DIFF
--- a/kong/error.go
+++ b/kong/error.go
@@ -1,13 +1,8 @@
 package kong
 
-import "fmt"
-
-type err404 struct {
-}
-
-func (e err404) Error() string {
-	return "Not found"
-}
+import (
+	"fmt"
+)
 
 type kongAPIError struct {
 	httpCode int
--- a/kong/error.go
+++ b/kong/error.go
@@ -1,5 +1,7 @@
 package kong
 
+import "fmt"
+
 type err404 struct {
 }
 
@@ -7,12 +9,22 @@ func (e err404) Error() string {
 	return "Not found"
 }
 
+type kongAPIError struct {
+	httpCode int
+	message  string
+}
+
+func (e *kongAPIError) Error() string {
+	return fmt.Sprintf("HTTP status %d (message: %q)", e.httpCode, e.message)
+}
+
 // IsNotFoundErr returns true if the error or it's cause is
 // a 404 response from Kong.
 func IsNotFoundErr(e error) bool {
-	if e == nil {
+	switch e := e.(type) {
+	case *kongAPIError:
+		return e.httpCode == 404
+	default:
 		return false
 	}
-	_, ok := e.(err404)
-	return ok
 }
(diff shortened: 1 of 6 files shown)


kong/response.go
```diff
@@ -16,15 +17,26 @@ func newResponse(res *http.Response) *Response {
 	return &Response{Response: res}
 }
 
+func messageFromBody(b []byte) string {
+	s := struct {
+		Message string
+	}{}
+
+	if err := json.Unmarshal(b, &s); err != nil {
+		return fmt.Sprintf("<failed to parse response body: %v>", err)
```
hbagdi, line 26: This `<>` is a new pattern in this codebase and I suggest to remove it (unless there is a reason for it that I don't know).

kong/response_test.go
```diff
@@ -0,0 +1,42 @@
...
+			name: "code 404",
+			response: http.Response{
+				StatusCode: 404,
+				Body:       ioutil.NopCloser(strings.NewReader(`{"message": "potayto pohtato", "some": "other field"}`)),
+			},
+			want: &kongAPIError{
+				httpCode: 404,
+				message:  "potayto pohtato",
+			},
+		},
```
hbagdi, line 35: Can we also add a test where json decoding fails?

kong/response.go
mflendrich: The reason for `<>` is that this string is a "meta" message which takes place of the actual message, where obtaining the actual message failed.

Compare the following two messages printed to the user:
1. `HTTP status 400 (message: failed to parse response body)`
2. `HTTP status 400 (message: <failed to parse response body>)`

To me, 1 suggests that "failed to parse response body" is a message from the server, instead of conveying that the message from the server was unreadable or nonexistent.

No strong preference here though. The above being said, I'm happy to solve this problem in any other reasonable way.

kong/response_test.go
mflendrich: Done!

kong/error.go
```diff
@@ -1,18 +1,25 @@
 package kong
 
-type err404 struct {
+import (
+	"fmt"
+)
+
+type kongAPIError struct {
+	httpCode int
+	message  string
```
hbagdi, line 9: This being unexported is good but we might be reaching a point where having an error type for this package is needed or having methods to retrieve further information from an error is needed.
This is simply a note for future development. Not changes required in this PR.

kong/response.go
hbagdi: Is that a standard? This is the first time I'm hearing about this interpretation of `<>`.

I'm a little conflicted here. On one hand, I prefer consistency so I'm against it but on the other hand this seems too small to worry about consistency. I'm fine either ways and my approval reflects that.

hbagdi (approved):

kong/response.go
mflendrich: > Is that a standard?

Python does that when you try to print something unprintable.

I too think it's a minor detail - let's rethink it if a need arises in the future.

PR #14, opened 2020-12-22
feat: add support for upstream node health

### LINKED ISSUE #12: Add support for the /upstreams/{name or id}/health/ route
Currently the library doesn't support to query the `/upstreams/{name or id}/health/` route (see [Show upstream health for node](https://docs.konghq.com/2.2.x/admin-api/#show-upstream-health-for-node) documentation for details).

It would be useful for controllers to have access to this route.

secat:
Closes #12


secat:
@hbagdi the pull request is ready for review. 

> **NOTE:** Only the  `enterprise-1.3.0.2` test fails

kong/types.go
```diff
@@ -146,6 +146,47 @@ type Healthcheck struct {
...
+}
+
+// HealthData represents the health data of a target
+// +k8s:deepcopy-gen=true
+type HealthData struct {
+	Host       *string             `json:"host,omitempty" yaml:"host,omitempty"`
+	Port       *int                `json:"port,omitempty" yaml:"port,omitempty"`
+	NodeWeight *int                `json:"nodeWeight,omitempty" yaml:"nodeWeight,omitempty"`
+	Weight     *HealthDataWeight   `json:"weight,omitempty" yaml:"weight,omitempty"`
+	Addresses  []HealthDataAddress `json:"addresses,omitempty" yaml:"addresses,omitempty"`
```
tjrivera, line 173: I think you'll want this type definition to refer to an array of pointers i.e `[]*HealthDataAddress`

kong/types.go
```diff
@@ -146,6 +146,47 @@ type Healthcheck struct {
...
+	DNS        *string             `json:"dns,omitempty" yaml:"dns,omitempty"`
+}
+
+// UpstreamNodeHealth represents the node health of a upstream
+// +k8s:deepcopy-gen=true
+type UpstreamNodeHealth struct {
+	ID        *string     `json:"id,omitempty" yaml:"id,omitempty"`
+	CreatedAt *float64    `json:"created_at,omitempty" yaml:"created_at,omitempty"`
```
tjrivera, line 181: time stamps are generally typed as `int` within this project. can we use that here as well?

kong/upstream_node_health_service.go
```diff
@@ -0,0 +1,61 @@
+package kong
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+// MTLSAuthService handles Upstream Node Healths in Kong.
```
tjrivera, line 9: Looks like a typo here `MTLSAuthService -> UpstreamNodeHealthService`

tjrivera (commented):
Nice addition. I made a first pass review of this PR. Just a few comments.

kong/types.go
secat: I was using at the beginning an `int`, however it is a `float64` that is returned.

By example:
```json
{
    "data": {
        "host":"10.0.0.1",
        "port":80,
        "weight":{
            "total":100,
            "available":100,
            "unavailable":0
        },
        "nodeWeight":100,
        "addresses":[
            {
                "port":80,
                "ip":"10.0.0.1",
                "health":"HEALTHCHECKS_OFF",
                "weight":100
            }
        ],
        "dns":"A"
    },
    "created_at":1608643189.157,
    "health":"HEALTHCHECKS_OFF",
    "id":"8f78d3dd-fcf3-4d9d-95d3-bce057ec13f5",
    "tags":null,
    "weight":100,
    "target":"10.0.0.1:80",
    "upstream":{
        "id":"f97522e0-2989-4dd2-8408-1eb390b65369"
    }
}
```

In this example, `created_at` has the value `1608643189.157` which is a float value.

Also it is inline with `Target` structure: https://github.com/Kong/go-kong/blob/main/kong/types.go#L172

kong/types.go
secat: Thanks for the review, updated the structure according to comment.

kong/upstream_node_health_service.go
secat: Good catch! Fixed the typo.

kong/types.go
tjrivera: Make sense! thanks :)

secat:
@hbagdi, @rainest my PR is ready for review. Thanks in advance.

rainest (approved):
We don't have anything that would use this beyond go-kong yet to try and test it further but the unit tests look fine and pass.

I had some confusion because the docs [are apparently wrong](https://github.com/Kong/go-kong/issues/12#issuecomment-772144524), but this looks correct for Kong's actual response.

secat:
I created the following issue https://github.com/Kong/kong/issues/6662 about the api documentation

PR #17, opened 2021-02-08
Use json.Number instead of float64

rainest:
Using float64 clashes with controller-gen's automatic manifest generation: https://github.com/Kong/railgun/issues/6

Review of the issues/PRs linked from controller-tools#245 shows most other projects changing to `string`. Doing this makes our unmarshalling unhappy. `json.Number` (still a [string](https://github.com/golang/go/blob/go1.15.8/src/encoding/json/decode.go#L188) under the hood) provides marshal/unmarshal helpers that avoid that. [resource.quantity](https://github.com/kubernetes/apimachinery/blob/v0.20.2/pkg/api/resource/quantity.go) from apimachinery, is an alternative option.

We currently use float64 for target creation timestamps and upstream health thresholds. Both are Lua numbers in Kong's schema (see https://github.com/Kong/kong/blob/2.3.1/kong/db/schema/entities/targets.lua#L22 and https://github.com/Kong/kong/blob/2.3.1/kong/db/schema/typedefs.lua#L309-L313 for the timestamps and https://github.com/Kong/kong/blob/2.3.1/kong/db/schema/entities/upstreams.lua#L66-L70 for thresholds).


shaneutt (approved):
Thanks for taking care of this, LGTM :partying_face:

kong/types.go
```diff
@@ -157,7 +157,7 @@ type PassiveHealthcheck struct {
 type Healthcheck struct {
```
mflendrich, line 157: This PR introduces changes of internal representation that are expected to preserve the interface. This needs to be tested.

Please implement tests checking that both JSON and YAML marshalling and unmarshalling correctly handle at least the following scenarios:
- field present
- field present but `0`
- field not present

kong/types.go
```diff
@@ -157,7 +157,7 @@ type PassiveHealthcheck struct {
...
 	Active    *ActiveHealthcheck  `json:"active,omitempty" yaml:"active,omitempty"`
 	Passive   *PassiveHealthcheck `json:"passive,omitempty" yaml:"passive,omitempty"`
-	Threshold *float64            `json:"threshold,omitempty" yaml:"threshold,omitempty"`
+	Threshold *json.Number        `json:"threshold,omitempty" yaml:"threshold,omitempty"`
```
mflendrich, line 160: Has it been verified that this change fixes the `make manifests` error in railgun?

mflendrich (changes requested):
LGTM in general, pending the (blocking) verifications in inline comments. :+1:


As an aside:
> [resource.quantity](https://github.com/kubernetes/apimachinery/blob/v0.20.2/pkg/api/resource/quantity.go) from apimachinery, is an alternative option

`go-kong`, being merely a golang binding for the Kong API, has nothing to do with Kubernetes on the inside. Adding a dependency to Kubernetes libraries would mean cross-wiring with an incidental user of this library that IMO wouldn't be justified here.

hbagdi:
-1.
KIC is the downstream project for library and will need to adapt accordingly. If controller runtime is doing something different, we should add in code in KIC to handle those differences.
This is a general purpose library that is used by multiple projects and a change like this will cause a lot of breakage. 
Please do not introduce any such changes into this library. If this was an overall positive change, this would be fine but AFAICS, there is only downside with this patch.

> We currently use float64 for target creation timestamps and upstream health thresholds. Both are Lua numbers in Kong's schema

If you look at Admin API responses, the returned timestamps have a higher granularity. Hence `float64`.

rainest:
> Adding a dependency to Kubernetes libraries would mean cross-wiring with an incidental user of this library that IMO wouldn't be justified here.

Right--to clarify, I wanted to just mention the options I'd seen. I think it'd be fine functionality-wise but adding the non-stdlib dependency here didn't seem like the right path.

@hbagdi can you help with what I'm missing/what breakage it'd cause? Intuitively, go-kong doesn't get to see a binary representation of these, since the admin API interactions are ultimately over text. For deck at least, this doesn't appear to cause issues:

```
11:57:30-0800 yagody $ git diff main                                                                                                                
diff --git a/go.mod b/go.mod
index 41579b9..13ecd3d 100644
--- a/go.mod
+++ b/go.mod
@@ -41,3 +41,5 @@ require (
 )
 
 go 1.13
+
+replace github.com/Kong/go-kong => github.com/Kong/go-kong v0.15.1-0.20210208191127-07e519dfcf17

11:58:00-0800 yagody $ go build -v -gcflags=all="-N -l" -ldflags "-X main.RELEASE=$TAG -X main.COMMIT=$COMMIT -X main.REPO=$REPO_INFO" -o /tmp/deckd

11:58:17-0800 yagody $ /tmp/deckd sync -s /tmp/float.yaml                                                                                           
creating upstream foo.upstream
Summary:
  Created: 1
  Updated: 0
  Deleted: 0
11:58:28-0800 yagody $ http localhost:8001/upstreams/foo.upstream | jq .healthchecks.threshold
50.33333
```
[float.yaml.txt](https://github.com/Kong/go-kong/files/5953956/float.yaml.txt)

And in the other direction:
```
curl -sX PATCH localhost:8001/upstreams/foo.upstream -H 'Content-Type: application/json' -d '{"healthchecks":{"threshold":20.2222}}'    
{"client_certificate":null,"created_at":1612900708,"id":"413131a1-2e62-4676-8b42-016e974908c9","tags":null,"name":"foo.upstream","algorithm":"round-robin","hash_on_header":null,"hash_fallback_header":null,"host_header":null,"hash_on_cookie":null,"healthchecks":{"threshold":20.2222,"active":{"unhealthy":{"http_statuses":[429,404,500,501,502,503,504,505],"tcp_failures":0,"timeouts":0,"http_failures":0,"interval":0},"type":"http","http_path":"\/","timeout":1,"healthy":{"successes":0,"interval":0,"http_statuses":[200,302]},"https_sni":null,"https_verify_certificate":true,"concurrency":10},"passive":{"unhealthy":{"http_failures":0,"http_statuses":[429,500,503],"tcp_failures":0,"timeouts":0},"healthy":{"http_statuses":[200,201,202,203,204,205,206,207,208,226,300,301,302,303,304,305,306,307,308],"successes":0},"type":"http"}},"hash_on_cookie_path":"\/","hash_on":"none","hash_fallback":"none","slots":10000}%                                                                                                     12:12:44-0800 yagody $ /tmp/deckd dump -o /tmp/patched.yaml                                                                                                
12:13:08-0800 yagody $ grep -i threshold /tmp/patched.yaml 
    threshold: 20.2222
```
Marshal/unmarshal are already automatically handled. It'd be a breaking change for libraries that want to make use of the float64 representation, but the type [has a standard function to convert it](https://golang.org/pkg/encoding/json/#Number.Float64).

go.mod
```diff
@@ -10,5 +10,6 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/satori/go.uuid v1.2.0
 	github.com/stretchr/testify v1.4.0
+	golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad // indirect
```
hbagdi, line 13: Why is this change necessary?
Does `go mod tidy` clean this up?

hbagdi:
Thanks for running those tests and sharing the examples.

Breaking change here has multiple meanings here:

The bigger and the most important point is that the motivation driving this is unacceptable. This library provides a relatively simple and idiomatic abstraction over Kong's API without any influence from k8s tooling (we do use k8s tooling to generate DeepCopy funcs but nothing in there is specific to k8s itself). Making changes which relax typing in Go code usually creates a problem in future. Types exist for a reason and we should adhere to them where possible. 

From a strict Go API perspective, changing a float64 to json.Number is a breaking change. Now, one can make the case that this is okay as per semantic versioning (this package is 0.x.y), but I think this unnecessary hassle to deal with for other users. It also sets a future problem that anytime a float64 is needed in the API, care must be taken to use a different data-type. It is hard to police the repo for future problems here.

Also, this is one of the "broken glass" things. We don't have anything like this in the codebase right now. If we make one exception to this, future us and new maintainers will use this patch to argue in favor of relaxing types

Lastly, while it works with decK, there are other users of this library. A good example here is https://github.com/kevholditch/terraform-provider-kong. There are other closed source code-bases that build on top of this as well and there is no way to exhaustively test all of them.

We could make an exception here but that would go against regular engineering practices.
Is this the only solution to the problem you are trying to solve or is there an alternative workaround that exists? If later, how expensive is the workaround? If former, has anyone else run into this issue in `controller-gen` and tried to fix it? Can we try to fix it upstream in that project?

mflendrich:
> The bigger and the most important point is that the motivation driving this is unacceptable.

This PR can be viewed as an indication of a potential mismatch between type systems in:
- Lua (and, by extension, Kong Admin API),
- go-kong,
- JSON,
- YAML.

Lua numbers are weakly typed (and a float64 internally).

JSON Schema has [`integer` and `number`](https://json-schema.org/understanding-json-schema/reference/numeric.html) where `number` has unspecified precision.

YAML has `integer` and `float`.

`go-kong` has golang `int`, `int64` and `float64`.

I think that, in order to resolve this situation, we need to answer the following question:

_In go-kong, Is the golang type system for numbers more appropriate than the JSON Schema type system?_

`go-kong` has been built with an assumption that the answer to this question is _yes_. I see this PR as a way of challenging this premise.

rainest:
> In go-kong, Is the golang type system for numbers more appropriate than the JSON Schema type system?
> go-kong has been built with an assumption that the answer to this question is yes. I see this PR as a way of challenging this premise.

IMO the risk that you mess up precision with the conversion back and forth is real in the absolute sense, though limited for the values likely in use for thresholds.

_Something_ has to be responsible for that, even if it's only Lua after it passes through the admin API. The rationale for using `json.Number` or some other string representation would be that if you need to perform mathematical computations on it in Go, you should be aware of that risk, perform the string<>float conversions yourself, and take appropriate action if needed.

Since the unsafe type option does exist in kubebuilder, however, we can table it for our purposes for now.

PR #19, opened 2021-02-26
fix: simplify codegen

hbagdi:
- k8s code-generator always generates code inside GOPATH.
  With this patch, this has been changed to use a temp directory.
- We now install deepcopy-gen only, this significantly speeds up code
  generation and verification.
- We no longer perform ugly checkouts of the code-generator repo and
  instead use go modules to track the tools we need.


hbagdi:
Enterprise test failures are unrelated.

hbagdi:
@shaneutt @rainest I think this has tripped both of you. No longer!

rainest (approved):

shaneutt (approved):

PR #21, opened 2021-02-27
chore: add enterprise 2.1 2.2 and 2.3 to test suite

hbagdi:


.github/workflows/test.yaml
```diff
@@ -53,9 +53,11 @@ jobs:
       - name: Run tests
         run: go test -v ./...
   test-enterprise:
+    # temporary to run all jobs and see all failures
+    continue-on-error: true
     strategy:
       matrix:
-        kong_version: ['enterprise-1.3.0.2', 'enterprise-1.5.0.2']
+        kong_version: ['enterprise-1.3.0.2', 'enterprise-1.5.0.9', 'enterprise-2.1.4.4', 'enterprise-2.2.1.0', 'enterprise-2.3.2.0']
```
mflendrich, line 60: Nit: maybe change the format from JSON array to bulleted list?

mflendrich (approved):

hbagdi:
So I spend a bit of time to setup CI for new enterprise versions and it seems we have some breakage in the API from Kong.
ping @rainest

rainest:
2.1 and onward use a new workspace implementation, which doesn't provide entity sharing at all:

https://github.com/Kong/kong-ee/pull/1794
https://github.com/Kong/kong-ee/commit/58c1c13d447ad48e83ec44826642e21557f2ee2e

The `/workspaces/WS_ID/entities/*` endpoints no longer exist and have no replacement, so [this test](https://github.com/Kong/go-kong/blob/v0.15.0/kong/workspace_service_test.go#L134-L192) should not run on 2.1+ at all: fdc494a

rainest:
Not sure what's up with the oauth2 failure on 2.1.4.4. That "all or none of these fields" message should come up only if the token DAO has a `mutually_required` rule, but it doesn't. Per https://github.com/Kong/kong-ee/commit/946568c7892d392e67a8bd185bfe17f18fd942c1 that field should be optional, and the longer history doesn't show any changes for that code since that'd have changed how it's validated.

hbagdi:
@rainest Thanks for digging into this.
Could you get some help from internal teams to figure out the cause for the Oauth test breakage. That seems like a breaking change and it is _potentially_ already impacting KIC.

ping @tjrivera. Did you use Workspace entities endpoint anywhere by any chance?

rainest:
Oauth test looks like it's something that you'd already run into previously: https://github.com/Kong/kong/issues/6173

The core bug remains open and likely won't be fixed any time soon.

I mistakenly thought this was something Enterprise-only and didn't look through the core history because that's sort of what tests reported, but it's actually because Enterprise release tags _look_ like semver pre-release tags, even though that's not how we're using them in practice.

As far as the semver range function is concerned, `2.1.4.0-enterprise < 2.1.0`, because pre-releases are less than final releases. As such, it passes the gate, whereas a 2.2.x.y-enterprise "pre-release" does not.

https://github.com/blang/semver/issues?q=is%3Aissue+pre shows others raising this question, but there doesn't appear to be any way around it/any way to ignore the pre-release during comparison. We'd need to chop off the Enterprise bits in `runWhenKong()` before comparing to avoid that problem completely. A less perfect approach is to change the version we compare against here:

```
diff --git a/kong/oauth2_auth_service_test.go b/kong/oauth2_auth_service_test.go
index 3326215..8c3404e 100644
--- a/kong/oauth2_auth_service_test.go
+++ b/kong/oauth2_auth_service_test.go
@@ -138,7 +138,7 @@ func TestOauth2CredentialGet(T *testing.T) {
 }
 
 func TestOauth2CredentialUpdate(T *testing.T) {
-       runWhenKong(T, "<2.1.0")
+       runWhenKong(T, "<=2.0.5")
        assert := assert.New(T)
 
```

tjrivera:
> ping @tjrivera. Did you use Workspace entities endpoint anywhere by any chance?

We did use the workspace entities endpoint to migrate a service and its associated routes, plugins, etc. from one workspace to another. We plan on dropping this as we move to Kong 2.x since we understand the workspace membership constructs have changed a bit in the new version.

PR #24, opened 2021-03-19
Add interface types for all go-kong services

rainest:
Add interface types for the various Kong endpoint services and a function to generate Kong API errors. This allows code using this library to define mock types that implement the service without interacting with a Kong instance and instead return canned responses for unit tests.

For example usage, see the WIP branches at
https://github.com/Kong/kubernetes-ingress-controller/tree/refactor/validator-mocking
https://github.com/Kong/kubernetes-ingress-controller/tree/refactor/validator-mocking-plugin


mflendrich (approved):
:+1:

PR #26, opened 2021-03-22
Developer roles handling

mmorel-35:
This pull request is for handling developer roles in Go-kong and finally to allow deck to do it.
I haven't included the permissions field accessible only on the detailed endpoint of the developer role

I cancelled the PR#25 and did a rebase on my main branch.


mmorel-35:
The failures seems to be caused by 401 http errors. I guess it's a matter of licence ?

rainest:
The CI environment lacks an Enterprise license at present, and while we'd like to add one, it's currently blocked due to other infrastructure changes that will affect how we handle development Enterprise licenses in general. I was able to test this against 2.3.2.0 successfully.

kong/credentials_service.go
```diff
@@ -8,8 +8,8 @@ import (
 	"reflect"
 )
 
-// abstractCredentialService handles credentials in Kong.
-type abstractCredentialService interface {
+// AbstractCredentialService handles credentials in Kong.
+type AbstractCredentialService interface {
```
rainest, line 12: What was the rationale for this and the related changes elsewhere? It looks unrelated to adding the developer role.

I believe the reason that this isn't exposed currently is that it's designed for use within the other credential types in the package, e.g. the BasicAuth service uses it, and that provides the public implementation for adding that specific credential type.

mmorel-35: I rollback this

PR #28, opened 2021-03-25
support List with pagination and list all for developer and rbac roles

mmorel-35:
This PR add support for pagination for developer and rbac roles.

PS : I don't know why it detects a change on every lines of the files I modified. It didn't happened on my previous PR.


mmorel-35:
Replaced by #29

PR #29, opened 2021-03-25
support List with pagination and list all for developer and rbac roles

mmorel-35:


rainest:
Ah, shoot--the changes here look generally correct, but changes in the tests (the check for 5 roles instead of 2 should have been failing before, but wasn't) made me realize that I'd missed something earlier when manually testing #26. The `runWhenEnterprise(T, ">=0.33.0", true)` means the test only runs if RBAC is enabled, which shouldn't matter for the Portal, so it should be `runWhenEnterprise(T, ">=0.33.0", false)`.

With those actually running, I realized that the `Create()` method for roles shouldn't actually use PUTs--the method isn't implemented and will result in a 405. Providing an ID in a POST body is valid, however, and does create a role with that ID.

The Portal isn't on by default either, and poses an additional challenge because simply enabling the Portal doesn't mean you can use the Portal, as it's enabled per-workspace, so tests need to set that up before they interact with entities. The workspace API allows that, but is a bit of an annoyance in manual test environments, since you can't quite clean it up after (the default files persist even if you disable it again).

rainest:
Continuing in #30 to include additional changes and work around CI not being able to run Enterprise tests on PRs from forks. @mmorel-35 I think we have everything we'd need from the fork branch, though if not probably we'll ask you for changes on #30 and cherry-pick them into the branch on this repo.

PR #41, opened 2021-04-08
Keep actions up to date with Dependabot

mmorel-35:
Configure github-actions update with dependabot


mflendrich (approved):
Looks cool, let's see how it works.

mmorel-35:
@mflendrich , it seems like the indentation is not right, my bad, do you need a fixing PR?

PR #44, opened 2021-04-15
craete or update

HieuQB:


mflendrich:
Hi @HieuQB,

The go-kong `Update` function uses the `PATCH` method, so using it to "create or update" is invalid according to the Kong Admin API spec.

Implementing "create or update" should be done by adding a `CreateOrUpdate` function utilizing the `PUT` method, as described in the Kong Admin API reference. This would need to be done for all types supported by go-kong as a cross-cutting effort, including appropriate testing.

PR #46, opened 2021-04-23
refactor: return a pointer to client

hbagdi:
This cosmetic change make it easier
for the caller to use the API.


mflendrich (approved):
Looks good.

PR #48, opened 2021-04-26
Expose Kong capabilities through a function in the client

mmorel-35:
Creates a function to expose kong's informations like it's version, if it's an enterprise edition , which database is used.
There is also a Credentials attribute to inform if they support tags which is not the case under 1.4.0 version.

As this is calculated here, deck won't require semver as a direct dependency to calculate it himself.

I also made the kong client aware of the workspace. This allows the Root to be / by default and /kong when a workspace is used.

I reused getSemVerVer of KIC as a version cleaner


kong/types.go
```diff
@@ -5,6 +5,16 @@ import (
 	"strings"
 )
 
+type Kong struct {
+	Version     string
+	Enterprise  bool
+	Database    string
+	Credentials struct {
+		hasTagSupport bool
+		minVersion    string
```
mmorel-35, line 14: I created this field to let people know which version is required to have tag support for credentials

rainest:
My initial thoughts:

- Adding `workspace` to the client builder is a significant breaking change. It will affect everything that uses this library, even though the required changes downstream are pretty small. We probably want to plan around when we release this as such. I don't think we can get the root behavior in here without it.
- I recommend keeping the semver type in place rather than converting it to a string. I don't think the dependency is a huge concern and any downstream version-gated behavior would need to use semver comparisons. Those will likely persist because not everything will have a go-kong.Kong capability entry and/or because a user may not want to upgrade to a newer go-kong version that adds one immediately.
- mtls-auth credentials are weird and didn't get tag support when other credentials did (the first version they support tags is `2.3.2.0-enterprise-edition`); we make use of this in the controller. We can add that in a follow-up.
- This has use case overlap with the [existing Enterprise feature enabled checks](https://github.com/Kong/go-kong/blob/v0.17.0/kong/kong_test.go#L158-L170) that we use in go-kong tests, but doesn't conflict with that implementation. We may want to figure out a way to unify that with this in the future.

kong/types.go
```diff
@@ -2,9 +2,22 @@ package kong
...
 	"strings"
 )
 
+type Kong struct {
+	Enterprise  bool
+	Portal      bool
+	Database    string
+	RBAC        string
+	Version     semver.Version
+	Credentials struct {
```
mmorel-35, line 15: The explicit name of the credentials could replace this part

mmorel-35:
* I have removed the workspace from the client constructor and used a setter to define it
* I kept Version as a semVer
* The structure has been completed with RBAC and Portal to use them on the part you quoted
* Concerning the tag support on each credentials, I'm open to suggestion on how to structure it.  There shall be the state and the minimum version that support it so the client can trace it in an error log if he sees that it's not supported.

kong/utils.go
```diff
@@ -2,8 +2,18 @@ package kong
...
+	"regexp"
 	"strings"
+
+	"github.com/blang/semver/v4"
+)
+
+var (
+	versionParts   = 4
+	kong140Version = semver.MustParse("1.4.0")
+	kong232Version = semver.MustParse("2.3.2-0")
```
mmorel-35, line 16: I had to use 2.3.2-0 instead of 2.3.2 because 2.3.2-0 is considered as a pre build of 2.3.2

rainest:
> I have removed the workspace from the client constructor and used a setter to define it

This doesn't quite get away from the breaking change, since client users would still need to remove any existing workspace path they're putting in the base URL and call `SetWorkspace()` for the root request to work properly. An alternative did occur to me though: the fallback logic approach originally used in decK is still viable with some adjustments. The client can determine whether it's in a workspace or not and get the root info without explicit configuration like so:

1. Attempt to request `GET /kong` first. If this returns a 200, the base URL includes a workspace.
2. If the initial request for `GET /kong` returns a 404, request `GET /`
3. If neither returns a successful response, error out.

One of those will work if the base URL is a Kong admin API. If we also need to know the workspace we're in, we can split the base URL path and use the last path segment.

> Concerning the tag support on each credentials, I'm open to suggestions on how to structure it. There shall be the state and the minimum version that support it so the client can trace it in an error log if he sees that it's not supported.

I should have been a bit clearer on this: mtls-auth is the only credential type that's like this due to some weird history in its development, so it needs a unique capability value. We shouldn't see any other such oddball credentials in the future; the rest can share a single capabilities value, and should for simplicity.

mmorel-35:
If we use the last path segment, a HEAD /workspaces/last_segment which returns a 200 should confirm the existence of the last_segment as a workspace?

Any idea for the naming of the field for tag support of the credentials which aren't mtlsauth ?

rainest:
> Any idea for the naming of the field for tag support of the credentials which aren't mtlsauth ?

No strong opinion really. The original

```
	Credentials struct {
		hasTagSupport bool
		minVersion    string
	}
```
is fine by me.

> If we use the last path segment, a HEAD /workspaces/last_segment which returns a 200 should confirm the existence of the last_segment as a workspace?

Looks like it, though the endpoint is a bit odd. If you're sending a request to a workspace other than the root/default workspace, those requests can see the workspace you're in, but not others:

```
$ http lHEAD ocalhost:8001/other/workspaces/default 
HTTP/1.1 404 Not Found

$ http HEAD localhost:8001/other/workspaces/other 
HTTP/1.1 200 OK
```
The root can see everything, so `localhost:8001/workspaces/other` is a 200.

hbagdi:
Hey @mmorel-35 
Can you share the motivation behind this change?

mmorel-35:
@hbagdi , this seems to be the most DRY way to expose the Version and the calculed tag supports as this is repeated between projects (deck and kic to start with). This way the client doesn't need to know the logic but just use boolean properties.

hbagdi:
> @hbagdi , this seems to be the most DRY way to expose the Version and the calculed tag supports as this is repeated between projects (deck and kic to start with). This way the client doesn't need to know the logic but just use boolean properties.

The versioning part could make sense. The tag part is going to go away in KIC and decK as they will drop support for Kong 1.x so we shouldn't be putting that in.

kong/utils.go
```diff
@@ -103,3 +113,51 @@ func HTTPClientWithHeaders(client *http.Client,
...
+		m[3] = strings.Replace(m[3], ".", "", -1)
+	}
+	v = fmt.Sprintf("%s.%s%s", m[1], m[2], m[3])
+	return semver.Make(v)
+}
+
+func getKong(root map[string]interface{}) (*Kong, error) {
+	version := root["version"].(string)
+	configuration := root["configuration"].(map[string]interface{})
+	semVer, err := cleanSemVer(version)
```
hbagdi, line 140: Let's not do this here since it is a lossy conversion.
The user of the library can interpret the version if needed.

kong/client.go
```diff
@@ -264,17 +265,36 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
...
 
-	var root map[string]interface{}
-	_, err = c.Do(ctx, req, &root)
+func (c *Client) Kong(ctx context.Context) (*Kong, error) {
+	root, err := c.Root(ctx)
 	if err != nil {
 		return nil, err
 	}
-	return root, nil
+	return getKong(root)
```
hbagdi, line 299: I don't think this pattern is good.
We should instead give a utility function to the user which takes the output `/` or `/kong` and answers the questions around support for tags, enterprise, etc.
The `kong` packages is intended for talking to the API. Interpreting these specifics should not be part of this package.

mmorel-35:
If I summarize, there is nothing left to take from that attempt. The PR can be be closed?

kong/utils.go
rainest: How about providing both the cleaned semver and the original string with all our extra content? Elsewhere we're likely going to be performing semver comparisons for the most part, and I think it makes sense to have a standard conversion rather than each user implementing their own `cleanSemVer()` variant.

hbagdi: I think a better approach would be:
```go
info, err := client.Root() // most powerful
version := kong.versionFromInfo(info) // package-level global function to fetch version from the info map
semanticVersion := utils.ParseSemanticVersion(version) // package-level function
```

This makes the code cleaner and keeps each function responsible for one and only one thing.

hbagdi:
@mmorel-35 Apologies for mixed reviews from our end on this one.
We met internally and here is what we think we should be doing here:
- We don't want to add capabilities feature just yet. Most of it won't be needed as we drop support for Kong 1.x series.
- We would like the API to be cleaner and here is how we want it to look like: https://github.com/Kong/go-kong/pull/48#discussion_r629524346

Again, apologies for the delay and confusion here. Let us know if you have questions.

mmorel-35:
@hbagdi , is this what you want ?

kong/client.go
```diff
@@ -101,6 +101,7 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "parsing URL")
 	}
+
```
hbagdi, line 104: Please remove this unrelated line jump.

kong/utils.go
```diff
@@ -103,3 +111,27 @@ func HTTPClientWithHeaders(client *http.Client,
 	}
 	return res
 }
+
+func ParseSemanticVersion(v string) (semver.Version, error) {
```
hbagdi, line 115: Please add a doc string.

kong/utils_test.go
```diff
@@ -125,4 +158,5 @@ func Test_requestWithHeaders(t *testing.T) {
 			}
 		})
 	}
+
```
hbagdi, line 161: Please remove this unrelated line jump.

kong/utils.go
```diff
@@ -103,3 +111,27 @@ func HTTPClientWithHeaders(client *http.Client,
...
+	}
+	if m[3] != "" {
+		m[3] = "-" + strings.Replace(m[3], "enterprise-edition", "enterprise", 1)
+		m[3] = strings.Replace(m[3], ".", "", -1)
+	}
+	v = fmt.Sprintf("%s.%s%s", m[1], m[2], m[3])
+	return semver.Make(v)
+}
+
+func versionFromInfo(info map[string]interface{}) string {
```
hbagdi, line 135: Please export this function so that it can be used by end users to extract version.
```suggestion
func VersionFromInfo(info map[string]interface{}) string {
```

kong/utils.go
```diff
@@ -2,8 +2,16 @@ package kong
...
 	"bytes"
+	"fmt"
 	"net/http"
+	"regexp"
 	"strings"
+
+	"github.com/blang/semver/v4"
+)
+
+var (
```
hbagdi, line 13:
```suggestion
const (
```

hbagdi (changes requested):

kong/utils.go
```diff
@@ -103,3 +111,30 @@ func HTTPClientWithHeaders(client *http.Client,
 	}
 	return res
 }
+
+//ParseSemanticVersion creates a semantic version from a string
+// fix enterprise edition semver adding patch number
+// fix enterprise edition version with dash
+// fix bad version formats like 0.13.0preview1
```
hbagdi, line 118:
```suggestion
// ParseSemanticVersion creates a semantic version from the version
// returned by Kong.
```

kong/utils.go
```diff
@@ -103,3 +111,30 @@ func HTTPClientWithHeaders(client *http.Client,
...
+	}
+	if m[3] != "" {
+		m[3] = "-" + strings.Replace(m[3], "enterprise-edition", "enterprise", 1)
+		m[3] = strings.Replace(m[3], ".", "", -1)
+	}
+	v = fmt.Sprintf("%s.%s%s", m[1], m[2], m[3])
+	return semver.Make(v)
+}
+
+//VersionFromInfo retrieves the version from the response of root
```
hbagdi, line 136:
```suggestion
// VersionFromInfo retrieves the version from the response of root
```

kong/utils.go
```diff
@@ -103,3 +111,30 @@ func HTTPClientWithHeaders(client *http.Client,
...
+// or /kong endpoints
+func VersionFromInfo(info map[string]interface{}) string {
+	return info["version"].(string)
```
hbagdi, line 139: This can panic if `version` doesn't exist in info.
Please use `v, ok := myMap["key"]` check here.

hbagdi (changes requested):

hbagdi:
@mmorel-35 We are still interested in adding `/` and `/kong` endpoints for looking up info. If you could send that in a separate PR, that would be totally cool.

hbagdi (approved):
Thank you for working on this!

PR #49, opened 2021-04-26
refactor: introduce a fast path when headers don't need to injected

hbagdi:


hbagdi:
CI failures are due to the brownout outage by Bintray.

rainest (approved):

PR #55, opened 2021-04-28
Update CI package locations and test versions

rainest:
Updates version lists and package locations to make CI work with our new package repository. Our repository has some structural differences from Bintray, and the new version gates in setup_kong.sh account for this.

Test runs (with some additional diagnostic modifications) can be seen on https://github.com/rainest/go-kong/pull/6

1.3.0.2 Enterprise will still fail with this change. Our current credentials are for Bintray, not download.konghq.com. However, download.konghq.com also uses basic auth, so I don't expect we'll need changes to the script, only to the repo secret.

### DIFF
--- a/.ci/setup_kong.sh
+++ b/.ci/setup_kong.sh
@@ -7,13 +7,19 @@ sudo apt-get update
 sudo apt-get install openssl libpcre3 procps perl wget zlibc
 
 function setup_kong(){
-  SWITCH="1.3.100"
+  SWITCH="1.3.000"
+  SWITCH2="2.0.000"
 
-  URL="https://kong.bintray.com/kong-deb/kong-${KONG_VERSION}.xenial.all.deb"
+  URL="https://download.konghq.com/gateway-1.x-ubuntu-xenial/pool/all/k/kong/kong_${KONG_VERSION}_all.deb"
 
   if [[ "$KONG_VERSION" > "$SWITCH" ]];
   then
-  URL="https://kong.bintray.com/kong-deb/kong-${KONG_VERSION}.xenial.amd64.deb"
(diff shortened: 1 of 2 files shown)


rainest:
Harry mentioned interest in migrating to some sort of Docker-based solution over installing deb packages. I reviewed options for this initially, but didn't see an obvious way to do it.

As best I can tell, the only way to run a background container in a workflow is using [a Job service](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices), similar to how we use Postgres. Services do not provide any means of overriding commands or running initialization tasks, and we need to somehow run `kong migrations bootstrap` before actually starting Kong.

Individual steps can [use Docker images and specify commands](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses), but these containers would not persist after the step (and if they did persist, they'd block other steps--I don't think there's a way to background the step and continue with later steps).

A job can [specify a container for the entire job](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer), but we'd no longer have access to the standard repo access, Go, and Postgres stuff we're currently using on the runner. We can probably work around that, but it's a bit more stuff to maintain.

Aside from all that, the switchover has an extra benefit: as download.konghq.com added public versions of the Ubuntu Enterprise repos, we can now run (most) Enterprise image tests against community PR, as they do not require the artifact download secret. However, those tests will run in free mode (the license secret is still unavailable), so Enterprise-specific functionality will still only run on local PRs. 1.3.0.2 is also not available in the public repo, and will fail on community PRs, but IMO it's unlikely that any new changes will actually break compatibility with 1.3.0.2 and not also break for 1.5.0.11 at this point.

shaneutt (approved):

PR #56, opened 2021-04-29
Check existence of tags or a workspace with the HEAD http verb instead of GET

mmorel-35:
This is to use a service instead of 

https://github.com/Kong/kubernetes-ingress-controller/blob/fa0afebe1ccd19551305c89008d829a10790a1a0/cli/ingress-controller/main.go#L293-L296

And

https://github.com/Kong/kubernetes-ingress-controller/blob/ed88ebee538681888f303c572c139410f4d686f8/cli/ingress-controller/util.go#L35-L42

The logic of HEAD testing shall be moved somewhere to be reuse but I don't see where for now

### DIFF
--- a/kong/workspace_service_test.go
+++ b/kong/workspace_service_test.go
@@ -32,6 +32,10 @@ func TestWorkspaceService(T *testing.T) {
 	assert.Nil(err)
 	assert.NotNil(workspace)
 
+	exists, err = client.Workspaces.Exists(defaultCtx, createdWorkspace.ID)
+	assert.Nil(err)
+	assert.True(*exists)
+
 	workspace.Comment = String("new comment")
 	workspace, err = client.Workspaces.Update(defaultCtx, workspace)
 	assert.Nil(err)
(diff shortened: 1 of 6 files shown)


rainest (approved):

kong/workspace_service_test.go
```diff
@@ -32,6 +32,10 @@ func TestWorkspaceService(T *testing.T) {
 	assert.Nil(err)
 	assert.NotNil(workspace)
 
+	exists, err = client.Workspaces.Exists(defaultCtx, createdWorkspace.ID)
+	assert.Nil(err)
+	assert.True(exists)
```
rainest, line 37: Jumped the gun on the initial review; realized that the Enterprise test failures were actually legit:
```
kong/workspace_service_test.go:35:2: undefined: exists
kong/workspace_service_test.go:37:14: undefined: exists
```
```suggestion
	var exists bool
	exists, err = client.Workspaces.Exists(defaultCtx, createdWorkspace.ID)
	assert.Nil(err)
	assert.True(exists)
```

rainest (changes requested):

rainest (approved):
Local branch CI run successful in https://github.com/Kong/go-kong/runs/2495253491

Issue #47, opened 2021-04-23
Drop fields and functions for Kong 1.x compatiblity


hbagdi:
This module is compatible with Kong 1.x series currently.
The scope of this issue is to find fields and endpoints that were removed in Kong 2.0 and drop those from the package.
One prominent example of this is `run_on` property on `Plugin` struct.

hbagdi:
https://github.com/Kong/go-kong/pull/58 drops the CI tests and moves this issue forward.

PR #58, opened 2021-05-04
chore(ci) drop testing against older versions of Kong

hbagdi:
1.x series is no longer supported and maintained by Kong Inc so removing
testing against those.

See #47

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,11 +22,6 @@ jobs:
     strategy:
       matrix:
         kong_version:
-        - '1.0.4'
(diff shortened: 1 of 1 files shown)


rainest:
Enterprise 1.3 and 1.5 are both out of full support, but do still have time left in sunset support: https://docs.konghq.com/enterprise/2.3.x/support-policy/#version-support-for-kong-gateway-enterprise

Is our policy for sunset support in go-kong effectively then that we won't test against them and will instead require an issue report from the field?

rainest (approved):

PR #62, opened 2021-05-13
Handle workspace in Kong's client

mmorel-35:
Fixes https://github.com/Kong/go-kong/pull/48#issuecomment-840122457

### DIFF
--- a/kong/client.go
+++ b/kong/client.go
@@ -9,6 +9,7 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -37,8 +38,9 @@ var (
 // Kong cluster
 type Client struct {
 	client                  *http.Client
-	rootURL                 string
-	workspace               string
+	defaultRootURL          string
+	workspace               string       // Do not access directly. Use Get/SetWorkspace().
+	workspaceLock           sync.RWMutex // Synchronizes access to workspace.
 	common                  service
 	Consumers               AbstractConsumerService
 	Developers              AbstractDeveloperService
@@ -118,7 +120,7 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "parsing URL")
 	}
-	kong.rootURL = url.String()
+	kong.defaultRootURL = url.String()
 
 	kong.common.client = kong
 	kong.Consumers = (*ConsumerService)(&kong.common)
@@ -165,32 +167,26 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	return kong, nil
 }
 
-// WithWorkspace return the client with a defined workspace
-func (c *Client) WithWorkspace(workspace string) *Client {
-	c.SetWorkspace(workspace)
-	return c
-}
-
 // SetWorkspace defines a workspace for the client
 func (c *Client) SetWorkspace(workspace string) {
+	c.workspaceLock.Lock()
+	defer c.workspaceLock.Unlock()
 	c.workspace = workspace
 }
 
-//ClearWorkspace clear the workspace
-func (c *Client) ClearWorkspace() {
-	c.workspace = ""
+//GetWorkspace return the workspace
+func (c *Client) GetWorkspace() string {
+	c.workspaceLock.RLock()
+	defer c.workspaceLock.RUnlock()
+	return c.workspace
 }
 
 // baseURL build the base URL from the rootURL and the workspace
-func (c *Client) baseURL() string {
-	if c.hasWorkspace() {
-		return c.rootURL + "/" + c.workspace
+func (c *Client) baseURL(workspace string) string {
+	if len(workspace) > 0 {
+		return c.defaultRootURL + "/" + workspace
 	}
-	return c.rootURL
-}
-
-func (c *Client) hasWorkspace() bool {
-	return len(c.workspace) > 0
+	return c.defaultRootURL
 }
 
 // Do executes a HTTP request and returns a response
@@ -312,17 +308,17 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
 // Root returns the response of GET request on root of Admin API (GET / or /kong with a workspace).
 func (c *Client) Root(ctx context.Context) (map[string]interface{}, error) {
 	endpoint := "/"
-	if c.hasWorkspace() {
+	if len(c.GetWorkspace()) > 0 {
 		endpoint = "/kong"
 	}
 	req, err := c.NewRequest("GET", endpoint, nil, nil)
 	if err != nil {
 		return nil, err
 	}
-	var informations map[string]interface{}
-	_, err = c.Do(ctx, req, &informations)
+	var info map[string]interface{}
+	_, err = c.Do(ctx, req, &info)
 	if err != nil {
 		return nil, err
 	}
-	return informations, nil
+	return info, nil
 }
--- a/kong/client.go
+++ b/kong/client.go
@@ -175,8 +175,8 @@ func (c *Client) SetWorkspace(workspace string) {
 	c.workspace = workspace
 }
 
-//GetWorkspace return the workspace
-func (c *Client) GetWorkspace() string {
+//Workspace return the workspace
+func (c *Client) Workspace() string {
 	c.workspaceLock.RLock()
 	defer c.workspaceLock.RUnlock()
 	return c.workspace
@@ -309,7 +309,7 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
 // Root returns the response of GET request on root of Admin API (GET / or /kong with a workspace).
 func (c *Client) Root(ctx context.Context) (map[string]interface{}, error) {
 	endpoint := "/"
-	if len(c.GetWorkspace()) > 0 {
+	if len(c.Workspace()) > 0 {
 		endpoint = "/kong"
 	}
 	req, err := c.NewRequest("GET", endpoint, nil, nil)
--- a/kong/client.go
+++ b/kong/client.go
@@ -165,10 +165,15 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	return kong, nil
 }
 
-// WithWorkspace defines a workspace for the client
+// WithWorkspace return the client with a defined workspace
 func (c *Client) WithWorkspace(workspace string) *Client {
-	c.workspace = workspace
+	c.SetWorkspace(workspace)
 	return c
+}
+
+// SetWorkspace defines a workspace for the client
+func (c *Client) SetWorkspace(workspace string) {
+	c.workspace = workspace
 }
 
 //ClearWorkspace clear the workspace
--- a/kong/client.go
+++ b/kong/client.go
@@ -172,16 +172,12 @@ func (c *Client) WithWorkspace(workspace string) *Client {
 }
 
 // baseURL build the base URL from the rootURL and the workspace
-func (c *Client) baseURL() (string, error) {
+func (c *Client) baseURL() string {
 	var baseURL = c.rootURL
 	if c.hasWorkspace() {
 		baseURL = baseURL + "/" + c.workspace
 	}
-	url, err := url.ParseRequestURI(baseURL)
-	if err != nil {
-		return "", errors.Wrap(err, "parsing URL")
-	}
-	return url.String(), nil
+	return baseURL
 }
 
 func (c *Client) hasWorkspace() bool {
--- a/kong/client.go
+++ b/kong/client.go
@@ -39,7 +39,7 @@ var (
 type Client struct {
 	client                  *http.Client
 	defaultRootURL          string
-	workspace               string       // Do not access directly. Use Get/SetWorkspace().
+	workspace               string       // Do not access directly. Use Workspace()/SetWorkspace().
 	workspaceLock           sync.RWMutex // Synchronizes access to workspace.
 	common                  service
 	Consumers               AbstractConsumerService
--- a/kong/client.go
+++ b/kong/client.go
@@ -37,7 +37,8 @@ var (
 // Kong cluster
 type Client struct {
 	client                  *http.Client
-	baseURL                 string
+	rootURL                 string
+	workspace               string
 	common                  service
 	Consumers               AbstractConsumerService
 	Developers              AbstractDeveloperService
@@ -117,7 +118,7 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "parsing URL")
 	}
-	kong.baseURL = url.String()
+	kong.rootURL = url.String()
 
 	kong.common.client = kong
 	kong.Consumers = (*ConsumerService)(&kong.common)
@@ -162,6 +163,29 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	}
 	kong.logger = os.Stderr
 	return kong, nil
+}
+
+// WithWorkspace defines a workspace for the client
+func (c *Client) WithWorkspace(workspace string) *Client {
+	c.workspace = workspace
+	return c
+}
+
+// baseURL build the base URL from the rootURL and the workspace
+func (c *Client) baseURL() (string, error) {
+	var baseURL = c.rootURL
+	if c.hasWorkspace() {
+		baseURL = baseURL + "/" + c.workspace
+	}
+	url, err := url.ParseRequestURI(baseURL)
+	if err != nil {
+		return "", errors.Wrap(err, "parsing URL")
+	}
+	return url.String(), nil
+}
+
+func (c *Client) hasWorkspace() bool {
+	return len(c.workspace) > 0
 }
 
 // Do executes a HTTP request and returns a response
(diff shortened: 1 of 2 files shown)


hbagdi:
@mmorel-35 I think the right approach here is to do the following that solves major problems in decK's codebase:
- introduce a `SetWorkspace(string)` and `ClearWorkspace()` method on `*kong.Client`; this internally recomputes `baseURL` for the client (we will need to make baseURL thread-safe as it was assumed to not change after client creation)
- Modify the behavior of `Root()` to use `/kong` whenever a workspace is set in the client
- Introduce a `Version()` method on the client which internally uses `Root()` and then sends back the version (without semver parsing)

These 3 could be 3 separate PRs that go in one after the other.
The first one here can greatly help improve decK's codebase as changing workspace context won't require instantiating new clients.

mmorel-35:
If you add a `SetWorkspace()` this can set an internal workspace field of the client. Then the NewRequest will be in charge of the construction of the URL with the baseURL and the workspace. There can be an internal function to check the presence of a workspace `hasWorkspace()`. It would then be used by Root to know which endpoint to call.
What would be the use for `ClearWorkspace()`?

kong/client.go
```diff
@@ -164,6 +165,34 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	return kong, nil
 }
 
+// WithWorkspace return the client with a defined workspace
+func (c *Client) WithWorkspace(workspace string) *Client {
+	c.SetWorkspace(workspace)
+	return c
+}
+
```
mflendrich, line 173: This function modifies the object it's being called on and, as such, is not different than just calling `SetWorkspace`. I would be misled into thinking that it creates a copy without modifying the receiver.

Why is this function needed? I think that implementing it the right way will likely be cumbersome. Suggestion: remove it altogether.

kong/client.go
```diff
@@ -164,6 +165,34 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
...
+// SetWorkspace defines a workspace for the client
+func (c *Client) SetWorkspace(workspace string) {
+	c.workspace = workspace
+}
```
mflendrich, line 177:
```suggestion
// SetWorkspace defines a workspace for the client
func (c *Client) SetWorkspace(workspace string) {
	c.workspaceLock.Lock()
	defer c.workspaceLock.Unlock()
	c.workspace = workspace
}
```

Also please implement `GetWorkspace` using `RLock()` and `RUnlock()` appropriately.

kong/client.go
```diff
@@ -164,6 +165,34 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
...
+
+//ClearWorkspace clear the workspace
+func (c *Client) ClearWorkspace() {
+	c.workspace = ""
```
mflendrich, line 181: This needs to be synchronized, so it might be best to use `SetWorkspace`:
```suggestion
	c.SetWorkspace("")
```

Or maybe we don't need this function at all, and `SetWorkspace("")` should be the public interface to clear workspace? @hbagdi @mmorel-35 do you see use for `ClearWorkspace` as opposed to `SetWorkspace("")`?

kong/client.go
```diff
@@ -164,6 +165,34 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
...
+	if c.hasWorkspace() {
+		return c.rootURL + "/" + c.workspace
+	}
+	return c.rootURL
+}
+
+func (c *Client) hasWorkspace() bool {
+	return len(c.workspace) > 0
+}
+
```
mflendrich, line 195:
```suggestion
```

This function poses a data race danger because it tempts the user to test `hasWorkspace` and then `GetWorkspace` without synchronization in between, and the data may change between these two calls.

kong/client.go
```diff
@@ -280,18 +309,20 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
...
-// Root returns the response of GET request on root of
-// Admin API (GET /).
+// Root returns the response of GET request on root of Admin API (GET / or /kong with a workspace).
 func (c *Client) Root(ctx context.Context) (map[string]interface{}, error) {
-	req, err := c.NewRequest("GET", "/", nil, nil)
+	endpoint := "/"
+	if c.hasWorkspace() {
+		endpoint = "/kong"
+	}
+	req, err := c.NewRequest("GET", endpoint, nil, nil)
```
mflendrich, line 318: There is a synchronization issue here that's not trivial to fix.

`NewRequest` internally makes a call to `baseURL` which locks the workspace for reading.
Before that, we're making a workspace reading operation to check whether the endpoint shall be `/` or `/kong`.
We cannot allow these two operations to happen between synchronization in between because the workspace can change after the first lock gets released and before the second gets acquired.
We cannot just put another big lock on top of these two because that can lead into a deadlock.

The best way out of this situation I can think of is to refactor `NewRequest` so that there's a subset thereof that doesn't perform locking internally, and use it here.

That would mean the following (not intentionally breaking) change to `NewRequest`:

```go
func (c *Client) NewRequestRaw(method, baseURL string, endpoint string, qs interface{},                                                                                                                                                                                                                      
  body interface{}) (*http.Request, error) {                                                                                                                                                                                                                                              
  if endpoint == "" {                                                                                                                                                                                                                                                                     
    return nil, errors.New("endpoint can't be nil")                                                                                                                                                                                                                                       
  }                                                                                                                                                                                                                                                                                       
  //body to be sent in JSON                                                                                                                                                                                                                                                               
  var buf []byte                                                                                                                                                                                                                                                                          
  if body != nil {                                                                                                                                                                                                                                                                        
    var err error                                                                                                                                                                                                                                                                         
    buf, err = json.Marshal(body)                                                                                                                                                                                                                                                         
    if err != nil {                                                                                                                                                                                                                                                                       
      return nil, err                                                                                                                                                                                                                                                                     
    }                                                                                                                                                                                                                                                                                     
  }                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                          
  //Create a new request                                                                                                                                                                                                                                                                  
  req, err := http.NewRequest(method, baseURL+endpoint,                                                                                                                                                                                                                               
    bytes.NewBuffer(buf))                                                                                                                                                                                                                                                                 
  if err != nil {                                                                                                                                                                                                                                                                         
    return nil, err                                                                                                                                                                                                                                                                       
  }                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                          
  // add body if needed                                                                                                                                                                                                                                                                   
  if body != nil {                                                                                                                                                                                                                                                                        
    req.Header.Add("Content-Type", "application/json")                                                                                                                                                                                                                                    
  }                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                          
  // add query string if any                                                                                                                                                                                                                                                              
  if qs != nil {                                                                                                                                                                                                                                                                          
    values, err := query.Values(qs)                                                                                                                                                                                                                                                       
    if err != nil {                                                                                                                                                                                                                                                                       
      return nil, err                                                                                                                                                                                                                                                                     
    }                                                                                                                                                                                                                                                                                     
    req.URL.RawQuery = values.Encode()                                                                                                                                                                                                                                                    
  }                                                                                                                                                                                                                                                                                       
  return req, nil                                                                                                                                                                                                                                                                         
}                        

func (c *Client) NewRequest(method, endpoint string, qs interface{},                                                                                                                                                                                                                      
  body interface{}) (*http.Request, error) {                                                                                                                                                                                                                                              
  return c.NewRequestRaw(method, c.baseURL(c.GetWorkspace()), endpoint, qs, body)
}                        

```

And this function calling `NewRequestRaw`, using the workspace that was read only once.

kong/client.go
```diff
@@ -164,6 +165,34 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
...
+	c.workspace = ""
+}
+
+// baseURL build the base URL from the rootURL and the workspace
+func (c *Client) baseURL() string {
+	if c.hasWorkspace() {
+		return c.rootURL + "/" + c.workspace
+	}
+	return c.rootURL
+}
```
mflendrich, line 190:
```suggestion
// baseURL build the base URL from the rootURL and the workspace
func (c *Client) baseURL(ws string) string {
	if len(ws) > 0 {
			return c.rootURL + "/" + ws
	}
	return c.rootURL
}
```

To avoid the need to read the workspace multiple times from different functions, accept the workspace name as a parameter, and expect the caller to read the workspace only once.

kong/client.go
```diff
@@ -280,18 +309,20 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
...
 
-// Root returns the response of GET request on root of
-// Admin API (GET /).
+// Root returns the response of GET request on root of Admin API (GET / or /kong with a workspace).
 func (c *Client) Root(ctx context.Context) (map[string]interface{}, error) {
-	req, err := c.NewRequest("GET", "/", nil, nil)
+	endpoint := "/"
+	if c.hasWorkspace() {
+		endpoint = "/kong"
+	}
```
mflendrich, line 317:
```suggestion
	ws := c.GetWorkspace()
	endpoint := "/"
	if len(ws) > 0 {
		endpoint = "/kong"
	}
```

kong/client.go
```diff
@@ -280,18 +309,20 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
...
+	req, err := c.NewRequest("GET", endpoint, nil, nil)
 	if err != nil {
 		return nil, err
 	}
-
-	var root map[string]interface{}
-	_, err = c.Do(ctx, req, &root)
+	var informations map[string]interface{}
```
mflendrich, line 322: Call this `info`:

https://jakubmarian.com/information-is-vs-informations-are-is-information-plural-or-singular-in-english/

mflendrich (changes requested):
@mmorel-35,

> If you add a `SetWorkspace()` this can set an internal workspace field of the client. Then the NewRequest will be in charge of the construction of the URL with the baseURL and the workspace.

See the attached comments - I've proposed a slightly different approach wrt. constructing the base URL, in order to avoid synchronization issues.

> There can be an internal function to check the presence of a workspace `hasWorkspace()`. It would then be used by Root to know which endpoint to call.

See the attached comments as well - we need to design the APIs so that we can ensure valid concurrent access to the workspace setting.

> What would be the use for `ClearWorkspace()`?

Funny legacy reasons. Some operations in decK require a client without a workspace, even if the tool has been run on a specific workspace. For sake of implementation in go-kong, let's just assume that the user of the client will have freedom to set, change and unset the workspace multiple times. But I believe that we could drop the `ClearWorkspace` idea in favor of `SetWorkspace("")` but that's not a strong opinion.

> About the `Version()` method, what about defining a `Informations` or `Kong` service which would hold it and also have the methods like `IsInMemory()`, `Database()`, `IsRBACEnabled()` or `IsPortalEnabled()` which are used between KIC and the go-kong's integration tests? 

On a high level I'd like us to have an API like this in go-kong, but that's probably out of scope of this PR. Could you please file an issue with a description of your idea, so that we can discuss it in a narrower context? go-kong would definitely benefit from such an improvement but let's pay due diligence before jumping into coding.

> Concerning thread-safe, how can it be insured ?

See my attached review comments.

kong/client.go
mmorel-35: Betrayed by my native language. Thank you for tip !

kong/client.go
```diff
@@ -164,6 +167,28 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
...
 
+// SetWorkspace defines a workspace for the client
+func (c *Client) SetWorkspace(workspace string) {
+	c.workspaceLock.Lock()
+	defer c.workspaceLock.Unlock()
+	c.workspace = workspace
+}
+
+//GetWorkspace return the workspace
+func (c *Client) GetWorkspace() string {
```
hbagdi, line 178: Idiomatic Go name for this function is `Workspace()`. Please update.
Let's wait for @mflendrich to respond before making the code change to avoid conflicting comments from me and Michal at the same time.

kong/client.go
```diff
@@ -164,6 +167,28 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	return kong, nil
 }
 
+// SetWorkspace defines a workspace for the client
```
hbagdi, line 170:
```suggestion
// SetWorkspace sets the Kong Enteprise workspace in the client.
// Calling this function with an empty string resets the workspace to default workspace.
```

kong/request.go
```diff
@@ -9,13 +9,9 @@ import (
 	"github.com/google/go-querystring/query"
 )
 
-// NewRequest creates a request based on the inputs.
-// endpoint should be relative to the baseURL specified during
-// client creation.
-// body is always marshaled into JSON.
-func (c *Client) NewRequest(method, endpoint string, qs interface{},
+//NewRequestRaw creates a request based on the inputs.
+func (c *Client) NewRequestRaw(method, baseURL string, endpoint string, qs interface{},
```
hbagdi, line 13: What if we rename baseURL to `url`? 
I _think_ it makes the signature a bit more clear.

kong/request.go
```diff
@@ -51,3 +46,12 @@ func (c *Client) NewRequest(method, endpoint string, qs interface{},
 	}
 	return req, nil
 }
+
+// NewRequest creates a request based on the inputs.
+// endpoint should be relative to the baseURL specified during
+// client creation.
+// body is always marshaled into JSON.
+func (c *Client) NewRequest(method, endpoint string, qs interface{},
+	body interface{}) (*http.Request, error) {
```
hbagdi, line 55: Is this a bit more descriptive?
```suggestion
func (c *Client) NewRequest(method, relativePath string, qs interface{},
	body interface{}) (*http.Request, error) {
```

hbagdi (changes requested):

kong/request.go
mmorel-35: Using url would mean using the full one, isn't it?
As it's always concatenated with the endpoint parameter, wouldn't it it lead to confusion?

kong/client.go
mmorel-35: It seems like it (https://golang.org/doc/effective_go#Getters), indeed. I'll wait.

mflendrich: Sounds good to me too :+1:

kong/client.go
```diff
@@ -37,7 +38,9 @@ var (
 // Kong cluster
 type Client struct {
 	client                  *http.Client
-	baseURL                 string
+	defaultRootURL          string
+	workspace               string       // Do not access directly. Use Get/SetWorkspace().
```
mflendrich: note: this comment needs to be updated when we get rid of the `GetWorkspace()` name

kong/client.go
```diff
@@ -280,18 +306,20 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
...
-// Root returns the response of GET request on root of
-// Admin API (GET /).
+// Root returns the response of GET request on root of Admin API (GET / or /kong with a workspace).
 func (c *Client) Root(ctx context.Context) (map[string]interface{}, error) {
-	req, err := c.NewRequest("GET", "/", nil, nil)
+	endpoint := "/"
+	if len(c.Workspace()) > 0 {
+		endpoint = "/kong"
+	}
+	req, err := c.NewRequest("GET", endpoint, nil, nil)
```
mflendrich, line 315:
```suggestion
	ws := c.Workspace()
	if len(ws) > 0 {
		endpoint = "/kong"
	}
	req, err := c.NewRequestRaw("GET", c.baseURL(ws), endpoint, nil, nil)
```

Like I said in a previous comment, we need to make sure that this function makes exactly one synchronized read of the workspace name. This suggestion is intended to achieve exactly that.

kong/request.go
hbagdi: I misunderstood. @mflendrich Should we name these as `NewWorkspacedRequest` `NewRequest` and then appropriate docs comment? The current API in this patch is not very easy to grok.

kong/client.go
```diff
@@ -164,6 +167,29 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
...
+
+//Workspace return the workspace
+func (c *Client) Workspace() string {
+	c.workspaceLock.RLock()
+	defer c.workspaceLock.RUnlock()
+	return c.workspace
+}
+
+// baseURL build the base URL from the rootURL and the workspace
+func (c *Client) baseURL(workspace string) string {
```
hbagdi, line 186:
```suggestion
func (c *Client) workspacedBaseURL(workspace string) string {
```
Makes it a little more descriptive and self-documenting.
@mflendrich Thoughts?

hbagdi:
ping @mflendrich

mflendrich (commented):
Seems that I reviewed this PR a week ago but the comments remained as pending :weary:. Sorry about that.

mflendrich (approved):
No blockers left from my side.

kong/client.go
mmorel-35: Done

hbagdi (approved):

mflendrich:
This may have something to do with a KIC 2.0 issue about the `/:workspace/kong` query URL being invalid: https://github.com/Kong/go-kong/pull/62

mmorel-35:
This PR has not been released yet, in v0.19.0 the root was just / not /kong, with a new release, and it's integration in kic, the issue https://github.com/Kong/kubernetes-ingress-controller/issues/1353 shall be fixed. It is what you meant @mflendrich, right?

PR #63, opened 2021-05-13
Use only int64 and float64

### LINKED ISSUE #54: Feature Request: Controller-Gen Compatibility
Currently in the [Kong Kubernetes Ingress Controller (KIC)][kic] we use [controller-gen][cg] as a generator for manifests and libraries within the project.

Currently types within this repository are not explicitly compatible due to Floats, and so the following flag has to be used:

```
allowDangerousTypes=true
```

The purpose of this issue is to ledger a request that we modify the types in a future `go-kong` release such that we would start _intentionally maintaining explicit Kubernetes compatibility_.

[kic]:https://github.com/kong/kubernetes-ingress-controller
[cg]:https://github.com/kubernetes/code-generator

mmorel-35:
Fixes #54

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,3 +15,6 @@ linters:
   - gomnd
   - maligned
   - errcheck
+linters-settings:
+  lll:
+    line-length: 150
--- a/kong/types.go
+++ b/kong/types.go
@@ -40,25 +40,26 @@ type CIDRPort struct {
 // Read https://getkong.org/docs/0.13.x/admin-api/#Route-object
 // +k8s:deepcopy-gen=true
 type Route struct {
-	CreatedAt               *int64              `json:"created_at,omitempty" yaml:"created_at,omitempty"`
-	Hosts                   []*string           `json:"hosts,omitempty" yaml:"hosts,omitempty"`
-	Headers                 map[string][]string `json:"headers,omitempty" yaml:"headers,omitempty"`
-	ID                      *string             `json:"id,omitempty" yaml:"id,omitempty"`
-	Name                    *string             `json:"name,omitempty" yaml:"name,omitempty"`
-	Methods                 []*string           `json:"methods,omitempty" yaml:"methods,omitempty"`
-	Paths                   []*string           `json:"paths,omitempty" yaml:"paths,omitempty"`
-	PathHandling            *string             `json:"path_handling,omitempty" yaml:"path_handling,omitempty"`
-	PreserveHost            *bool               `json:"preserve_host,omitempty" yaml:"preserve_host,omitempty"`
-	Protocols               []*string           `json:"protocols,omitempty" yaml:"protocols,omitempty"`
-	RegexPriority           *int64              `json:"regex_priority,omitempty" yaml:"regex_priority,omitempty"`
-	Service                 *Service            `json:"service,omitempty" yaml:"service,omitempty"`
-	StripPath               *bool               `json:"strip_path,omitempty" yaml:"strip_path,omitempty"`
-	UpdatedAt               *int64              `json:"updated_at,omitempty" yaml:"updated_at,omitempty"`
-	SNIs                    []*string           `json:"snis,omitempty" yaml:"snis,omitempty"`
-	Sources                 []*CIDRPort         `json:"sources,omitempty" yaml:"sources,omitempty"`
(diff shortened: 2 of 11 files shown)


mmorel-35:
I have used float on created At and updatedAt because of this : https://github.com/Kong/go-kong/runs/2574465387?check_suite_focus=true#step:7:279

.golangci.yml
```diff
@@ -15,3 +15,6 @@ linters:
   - gomnd
   - maligned
   - errcheck
+linters-settings:
+  lll:
+    line-length: 150
```
mflendrich, line 20: `lll` has 120 as the default and this is what we have configured in other repos. Why do you want to change that?

Suggestion: remove the customization (leave the default) and wrap all changed lines that require it.

mflendrich (commented):
Looks good, please however:
- see the comment https://github.com/Kong/go-kong/issues/54#issuecomment-840673139 - it doesn't affect the code of this PR, just FYI
- see the action required (in the attached comment)
- be advised that I'm checking with maintainers of Kong Gateway that the `created_at` change from int to float is appropriate.

kong/types.go
```diff
@@ -10,56 +10,55 @@ import (
...
+	Protocols               []*string           `json:"protocols,omitempty" yaml:"protocols,omitempty"`
+	RegexPriority           *int64              `json:"regex_priority,omitempty" yaml:"regex_priority,omitempty"`
+	Service                 *Service            `json:"service,omitempty" yaml:"service,omitempty"`
+	StripPath               *bool               `json:"strip_path,omitempty" yaml:"strip_path,omitempty"`
+	UpdatedAt               *float64            `json:"updated_at,omitempty" yaml:"updated_at,omitempty"`
+	SNIs                    []*string           `json:"snis,omitempty" yaml:"snis,omitempty"`
+	Sources                 []*CIDRPort         `json:"sources,omitempty" yaml:"sources,omitempty"`
+	Destinations            []*CIDRPort         `json:"destinations,omitempty" yaml:"destinations,omitempty"`
+	Tags                    []*string           `json:"tags,omitempty" yaml:"tags,omitempty"`
+	HTTPSRedirectStatusCode *int64              `json:"https_redirect_status_code,omitempty" yaml:"https_redirect_status_code,omitempty"`
```
mmorel-35, line 61: @mflendrich, this line is the reason why I changed lll parameters.
I can't find a way to resize it beneath 121 characters with int it's fine but with int64, it isn't

rainest: golangci supports per-file exemptions, which would be appropriate here. go-kong doesn't currently have any, but deck does: https://github.com/Kong/kong/blob/master/kong.conf.default#L98-L126 is an example for this rule.

rainest:
I don't think this fixes the issue. https://github.com/kubernetes-sigs/controller-tools/issues/245 and the [controller-gen source](https://github.com/kubernetes-sigs/controller-tools/blob/9cd8c2840e842554b3264634e3956e9f4d1c5497/pkg/crd/schema.go#L403-L433) indicates that `float64` isn't any permissible than `float`. The restriction exists because float conversions are inconsistent in and out of strings in JSON, not because of a need to support large or higher-precision values.

We'd approached this before in https://github.com/Kong/go-kong/pull/17, which moves to a string-backed type, but decided against it because it's a breaking change in go-kong and `allowDangerousTypes` is available for KIC. The dangerous conversion should be minimal impact, since the problem fields are things we ignore in practice (created/updated at timestamps) or values where precision weirdness is probably low impact (health thresholds).

mmorel-35:
OK, we can drop the PR so ?

kong/types.go
rainest: Nope. Clipboard hurray. https://github.com/Kong/deck/blob/6fd816b2bb4603dc2479beeff8b6e1e7a6e1e167/.golangci.yml#L37-L40 was what I meant.

rainest:
Yeah--wouldn't help the issue with controller-gen and we don't have other reasons to modify the types.

FWIW, the odd-looking discrepancy in the timestamp types is intentional. Kong actually defines both [integer and float timestamp types](https://github.com/Kong/kong/blob/2.4.0/kong/db/schema/typedefs.lua#L331-L342), though only targets use the float variant.

PR #65, opened 2021-05-15
Defines the Information Service on top of Kong's Root endpoint

mmorel-35:
@hbagdi, after your comment  https://github.com/Kong/go-kong/pull/62#issuecomment-841565681 , here is my suggestion.


mflendrich (changes requested):
The proposed API design:
- makes a new HTTP request for each separate field of information
- treats the returned object as unstructured

We should instead:
- define a data structure to deserialize the JSON object into
- have just one API on the service that pulls the kong info endpoint (`/` or `/:workspace/kong`)
- define accessor methods for individual bits of information on that object, instead of implementing them on the abstract service

mmorel-35:
What do you think of something like this :
```
type Info struct {
    Version string `json:version`
    Configuration struct {
        Database... 
        Portal... 
        RBAC...
   }
}

func (p *Info) IsInMemory()
func (p *Info) IsEnterprise()
```

kong/information.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type Info struct {
+	Version       string `json:"version,omitempty" yaml:"version,omitempty"`
+	Configuration struct {
+		Database string `json:"database,omitempty" yaml:"database,omitempty"`
+		Portal   bool   `json:"portal,omitempty" yaml:"portal,omitempty"`
+		RBAC     string `json:"rbac,omitempty" yaml:"rbac,omitempty"`
+	} `json:"configuration,omitempty" yaml:"configuration,omitempty"`
+}
```
mflendrich, line 16: This type definition needs to go to `kong/types.go` for consistency with other API data types.

kong/information.go
```diff
@@ -0,0 +1,53 @@
...
+
+// IsRBACEnabled check if RBAC are enabled
+func (s *Info) IsRBACEnabled() bool {
+	return "on" == s.Configuration.RBAC
+}
+
+// IsPortalEnabled check if the portal is enabled
+func (s *Info) IsPortalEnabled() bool {
+	return s.Configuration.Portal
+}
```
mflendrich, line 36: `s.IsPortalEnabled()` is equivalent to `s.Configuration.Portal`. There's no need for this function to exist if the user can just read the bool from the public field.

Suggestion: remove

kong/information.go
```diff
@@ -0,0 +1,53 @@
...
+		Database string `json:"database,omitempty" yaml:"database,omitempty"`
+		Portal   bool   `json:"portal,omitempty" yaml:"portal,omitempty"`
+		RBAC     string `json:"rbac,omitempty" yaml:"rbac,omitempty"`
+	} `json:"configuration,omitempty" yaml:"configuration,omitempty"`
+}
+
+// IsEnterprise check if Kong is enterprise edition
+func (s *Info) IsEnterprise() bool {
+	return strings.Contains(s.Version, "enterprise")
+}
```
mflendrich, line 21: Interpreting the version is a part of business logic that does not belong in the API binding.

Suggestion: remove

kong/information.go
```diff
@@ -0,0 +1,53 @@
...
+
+// IsInMemory check if Kong is in memory
+func (s *Info) IsInMemory() bool {
+	return "off" == s.Configuration.Database
+}
+
+// IsRBACEnabled check if RBAC are enabled
+func (s *Info) IsRBACEnabled() bool {
+	return "on" == s.Configuration.RBAC
+}
```
mflendrich, line 31: Please implement these on the `Configuration` object directly. (Will probably need un-inlining the `Configuration` `struct` definition).

kong/information.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	bytes, err := json.Marshal(information)
+	if err != nil {
+		return nil, err
+	}
+	var info Info
+	err = json.Unmarshal(bytes, &info)
+	if err != nil {
+		return nil, err
+	}
```
mflendrich, line 51: Please implement this as a function:

```go
func Convert(from, to interface{}) error {
  bytes, err := json.Marshal(from)
  if err != nil {
    return err
  }
  return json.Unmarshal(bytes, to)
}

and unit-test it.
```

kong/information.go
```diff
@@ -0,0 +1,53 @@
...
+func (s *Info) IsRBACEnabled() bool {
+	return "on" == s.Configuration.RBAC
+}
+
+// IsPortalEnabled check if the portal is enabled
+func (s *Info) IsPortalEnabled() bool {
+	return s.Configuration.Portal
+}
+
+func (c *Client) Info(ctx context.Context) (*Info, error) {
```
mflendrich, line 38: > And as it would be a single function it doesn't need an entire service of it's own. So it would directly be at the client's level.

There are 3 problems I can think of with this approach:
1. Implementing directly at client's level prevents mocking the info service, and requires the users to operate on a very concrete type with lots of responsibility (remember that `type Client` has all the services as its fields). Better to implement a narrow type that allows querying about info.
2. Once the Admin API has an OpenAPI spec (which can happen soon) we'll be able to replace Kong Admin API struct definitions with autogenerated ones. Making the info service consistent will simplify that process.
3. The root endpoint is a resource like any other, and we should be consisitent with the other getters/setters for other resources.

Therefore please implement an `InfoService` with only the `Get` method on it. I agree that it's more code in this instance, but, in my opinion, the benefits outweigh the cost.

kong/information.go
mmorel-35: There is a conflict with another Configuration, here
https://github.com/Kong/go-kong/blob/402cf0f5dd1df3f56059fb3f152bbd1c938b6a84/kong/types.go#L237

What do you suggest?

mflendrich: Let's go with `RuntimeConfiguration`.

kong/information_service.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// IsRBACEnabled check if RBAC are enabled
+func (r *RuntimeConfiguration) IsRBACEnabled() bool {
+	return "on" == r.RBAC
+}
+
+// Convert convert a object to another through json marshalling
+// unmarshalling
+func Convert(from, to interface{}) error {
```
hbagdi, line 28: Do we really need this to be part of the public API?

kong/types.go
```diff
@@ -438,3 +438,18 @@ type DeveloperRole struct {
...
+	Configuration *RuntimeConfiguration `json:"configuration,omitempty" yaml:"configuration,omitempty"`
+}
+
+// RuntimeConfiguration represents the runtime configuration of Kong.
+// +k8s:deepcopy-gen=true
+type RuntimeConfiguration struct {
+	Database string `json:"database,omitempty" yaml:"database,omitempty"`
+	Portal   bool   `json:"portal,omitempty" yaml:"portal,omitempty"`
+	RBAC     string `json:"rbac,omitempty" yaml:"rbac,omitempty"`
+}
```
hbagdi, line 455: @mflendrich Should we consider simplifying this by moving the 3 fields in the RuntimeConfiguration struct to Info struct?
The extra struct seems unnecessary to me.

kong/information_service.go
mflendrich: You're right - my comment https://github.com/Kong/go-kong/pull/65#discussion_r636800479 was misleading in this bit.

@mmorel-35 please unexport this function by renaming it to `convert`, and sorry about the confusion.

kong/types.go
mflendrich: Is there a way to do that without breaking JSON/YAML marshaling? If so, :100:. But, judging by the JSON marshaler spec, I don't know how to do that without an intermediate `struct`.

kong/information_service.go
```diff
@@ -0,0 +1,48 @@
+package kong
+
+import (
+	"context"
+	"encoding/json"
+)
+
+// AbstractInfoService handles Kong's Information.
+type AbstractInfoService interface {
+	// Get retrieves the information concerning Kong
```
mflendrich, line 10:
```suggestion
	// Get retrieves the general runtime information about the Kong gateway.
```

kong/information_service.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Get retrieves the information concerning Kong
+func (s *InfoService) Get(ctx context.Context) (*Info, error) {
+	information, err := s.client.Root(ctx)
+	if err != nil {
+		return nil, err
+	}
+	var info Info
+	err = convert(information, &info)
+	if err != nil {
```
mflendrich, line 44:
```suggestion
	if err := convert(information, &info); err != nil {
```

kong/types.go
```diff
@@ -438,3 +438,18 @@ type DeveloperRole struct {
...
+
+// Info represents the information concerning Kong.
+// +k8s:deepcopy-gen=true
+type Info struct {
+	Version       string                `json:"version,omitempty" yaml:"version,omitempty"`
+	Configuration *RuntimeConfiguration `json:"configuration,omitempty" yaml:"configuration,omitempty"`
+}
+
+// RuntimeConfiguration represents the runtime configuration of Kong.
+// +k8s:deepcopy-gen=true
```
mflendrich, line 450: Do we need `deepcopy` for this type?

kong/information_service_test.go
```diff
@@ -0,0 +1,32 @@
+package kong
```
mflendrich, line 1: This repo has simple integration tests for other services.

See the `service_service_test.go` for reference where it uses `NewTestClient()` and asserts about the responses received therefrom.

Please at least check that Get() returns no error and that the version and database are not nil in the returned message.

kong/information_service_test.go
```diff
@@ -0,0 +1,32 @@
...
+		Version: "2.3.3.2-enterprise-edition",
+		Configuration: &RuntimeConfiguration{
+			Portal:   true,
+			RBAC:     "on",
+			Database: "postgres",
+		},
+	}
+	var actual Info
+	err := convert(information, &actual)
+	assert.Nil(err)
```
mflendrich, line 30:
```suggestion
	assert.NoError(err)
```

mflendrich (commented):
Some minor style issues + a request for a simple integration test. Otherwise the information service looks good to me.

Please consider this comment https://github.com/Kong/go-kong/pull/65#discussion_r637255239 non-blocking - we'll make adjustments before merging if necessary.

kong/types.go
mflendrich: It seems that we don't. Please remove the deepcopy annotation from this type.

kong/types.go
mflendrich: This can probably be resolved async in a follow-up PR if necessary - that looks like a minor style issue. @hbagdi please speak up otherwise.

mflendrich (commented):
Please remove the deepcopy bit (as requested in the attached comment) and this PR is good to go.

hbagdi:
Deferring the entire review to @mflendrich here.

mmorel-35:
Hi @mflendrich, anything left to do here?

kong/information_service.go
```diff
@@ -0,0 +1,47 @@
...
+// unmarshalling
+func convert(from, to interface{}) error {
+	bytes, err := json.Marshal(from)
+	if err != nil {
+		return err
+	}
+	return json.Unmarshal(bytes, to)
+}
+
+// Get retrieves the information concerning Kong
```
mflendrich, line 36:
```suggestion
// Get retrieves the high-level metadata of a Kong instance.
```

kong/types.go
mflendrich: Closing as (apparently) a minor style issue.

mflendrich (approved):
Looks good now!

PR #66, opened 2021-05-18
refactor: simplify error creation and wrapping

hbagdi:
Since this package predates Go 1.13 which introduced error wrapping using `%w`
qualifier. This patch does the following:
- replaces `errors.Wrap` with `fmt.Errorf`
- remove `github.com/pkg/errors` as a dependency
- removes use of `errors.New` stdlib package with `fmt.Errorf`

### DIFF
--- a/kong/custom/entity_crud.go
+++ b/kong/custom/entity_crud.go
@@ -1,7 +1,7 @@
 package custom
 
 import (
-	"errors"
+	"fmt"
 	"regexp"
 	"strings"
 )
@@ -53,7 +53,7 @@ func render(template string, entity Entity) (string, error) {
 		if v := entity.GetRelation(m[1]); v != "" {
 			result = strings.Replace(result, m[0], v, 1)
 		} else {
-			return "", errors.New("cannot substitute '" + m[1] +
+			return "", fmt.Errorf("cannot substitute '" + m[1] +
 				"' in URL: " + template)
 		}
 	}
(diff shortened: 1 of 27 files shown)


kong/custom/entity_crud.go
```diff
@@ -53,7 +53,7 @@ func render(template string, entity Entity) (string, error) {
 		if v := entity.GetRelation(m[1]); v != "" {
 			result = strings.Replace(result, m[0], v, 1)
 		} else {
-			return "", errors.New("cannot substitute '" + m[1] +
+			return "", fmt.Errorf("cannot substitute '" + m[1] +
```
mmorel-35, line 56: Would it be an opportunity to use verbs here?

kong/custom_entity_service.go
```diff
@@ -35,7 +35,7 @@ func (s *CustomEntityService) Get(ctx context.Context,
 	entity custom.Entity) (custom.Entity, error) {
 	def := s.client.Lookup(entity.Type())
 	if def == nil {
-		return nil, errors.New("entity '" + string(entity.Type()) +
+		return nil, fmt.Errorf("entity '" + string(entity.Type()) +
```
mmorel-35, line 38: Would it be an opportunity to use verbs here and in the future rest of the file ?

rainest (approved):

kong/custom/entity_crud.go
hbagdi: @mmorel-35 Mind sending a follow up PR to keep the scope small here?

PR #68, opened 2021-05-27
Define FriendlyName for the entities, based on deck/state/types.go

mmorel-35:
This feature add identifier methods to entities to allow to print names or ids depending on which one is available.
It will offer a clean solution for https://github.com/Kong/deck/issues/417


kong/identifiers.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return *r1.ID
+}
+
+// Identifier returns a composite ID base on Role ID, workspace, and endpoint
+func (r1 *RBACEndpointPermission) Identifier() string {
+	if r1.Endpoint != nil {
+		return fmt.Sprintf("%s-%s-%s", *r1.Role.ID, *r1.Workspace, *r1.Endpoint)
+	}
+	return *r1.Endpoint
```
mflendrich, line 90: `Endpoint` is nil here and it cannot be dereferenced.

kong/identifiers.go
```diff
@@ -0,0 +1,91 @@
+package kong
+
+import "fmt"
+
+// Identifier returns the endpoint key name or ID.
+func (s1 *Service) Identifier() string {
```
mflendrich, line 6: Please call this function `FriendlyName` to better describe its purpose.

kong/identifiers.go
```diff
@@ -0,0 +1,91 @@
+package kong
```
mflendrich, line 1: Please implement methods alongside type definitions, in the same file.

kong/identifiers.go
```diff
@@ -0,0 +1,91 @@
...
+
+import "fmt"
+
+// Identifier returns the endpoint key name or ID.
+func (s1 *Service) Identifier() string {
+	if s1.Name != nil {
+		return *s1.Name
+	}
+	return *s1.ID
```
mflendrich, line 10: For all types please only dereference ID if it's non-nil.

If none of the checked fields is non-nil, return `""` as the friendly name.

kong/identifiers.go
```diff
@@ -0,0 +1,91 @@
...
+func (r1 *RBACRole) Identifier() string {
+	if r1.Name != nil {
+		return *r1.Name
+	}
+	return *r1.ID
+}
+
+// Identifier returns a composite ID base on Role ID, workspace, and endpoint
+func (r1 *RBACEndpointPermission) Identifier() string {
+	if r1.Endpoint != nil {
```
mflendrich, line 87: This needs to check all the dereferenced fields.

kong/identifiers.go
mflendrich, line 6: Please drop the `1`s from all the receiver names.

mmorel-35: There is already Identifier or Console which behave the same way in deck, are you sure about this name?

mmorel-35: It does, I'll take care of this this weekend

mmorel-35:
Hi @mflendrich, anything left to do here?

mflendrich (approved):

mmorel-35:
Any idea when is planned the next release for go-kong?

PR #70, opened 2021-05-29
Update Golangci to v1.40.1 and fix lint 

mmorel-35:


mflendrich (approved):
Thank you! :pray: :100:

PR #75, opened 2021-07-15
Use github.com/google/uuid instead of github.com/satori/go.uuid

mmorel-35:
This PR is for avoiding to have several dependencies responsible for the uuid handling. It will reduce the number of dependencies in deck and kic and ktf


shaneutt (approved):
LGTM :+1: 

Thanks for following up on this from the PR where it was originally brought up :bow:

PR #89, opened 2021-09-21
change workspace.Exists() endpoints to /{ws}/workspaces/{ws}

ccfishk:
change workspace.Exists() endpoints from /workspaces endpoints to /{ws}/workspaces/{ws} require less permission

### DIFF
--- a/kong/exists.go
+++ b/kong/exists.go
@@ -8,7 +8,11 @@ import (
 // exists check the existence  with a HEAD HTTP verb
 func (c *Client) exists(ctx context.Context,
 	endpoint string) (bool, error) {
-	req, err := c.NewRequest("HEAD", endpoint, nil, nil)
+	// Originally, this used HEAD. We promptly discovered that HEAD doesn't actually work for this
+	// in Kong <=2.6: https://github.com/Kong/kong/issues/7554
+	// Although future versions will support HEAD for existence checks, using GET for backwards
+	// compatibility at the cost of efficiency
+	req, err := c.NewRequest("GET", endpoint, nil, nil)
 	if err != nil {
 		return false, err
 	}
--- a/kong/workspace_service.go
+++ b/kong/workspace_service.go
@@ -49,7 +49,7 @@ func (s *WorkspaceService) Exists(ctx context.Context,
 		return false, fmt.Errorf("nameOrID cannot be nil for Get operation")
 	}
 
-	endpoint := fmt.Sprintf("/workspaces/%v", *nameOrID)
+	endpoint := fmt.Sprintf("/%v/workspaces/%v", *nameOrID, *nameOrID)
 	return s.client.exists(ctx, endpoint)
 }
 


rainest:
Ah, okay, my solution doesn't quite work with what this allows. You can only use the workspace name as the prefix, so this doesn't work if you try to use an ID, which is what the test checks:

```
> github.com/kong/go-kong/kong.(*WorkspaceService).Exists() ./workspace_service.go:53 (PC: 0x7f8225)
    48:		if isEmptyString(nameOrID) {
    49:			return false, fmt.Errorf("nameOrID cannot be nil for Get operation")
    50:		}
    51:	
    52:		endpoint := fmt.Sprintf("/%v/workspaces/%v", *nameOrID, *nameOrID)
=>  53:		return s.client.exists(ctx, endpoint)
    54:	}
    55:	
    56:	// Create creates a Workspace in Kong.
    57:	func (s *WorkspaceService) Create(ctx context.Context,
    58:		workspace *Workspace) (*Workspace, error) {
(dlv) p endpoint
"/95b0679f-467c-4374-a161-7e639cba7e91/workspaces/95b0679f-467c-4374-a161-7e639cba7e91"
```
```
$ http localhost:8001/95b0679f-467c-4374-a161-7e639cba7e91/workspaces/95b0679f-467c-4374-a161-7e639cba7e91 
HTTP/1.1 404 Not Found
$ http localhost:8001/teamA/workspaces/95b0679f-467c-4374-a161-7e639cba7e91
HTTP/1.1 200 OK
```
We can't differentiate names from IDs without fuzzy logic (strings that look parse as UUIDs are _probably_ IDs, but it's not guaranteed) or a breaking signature change to explicitly indicate what field you're requesting, which we don't want.

Alternatively this could use a fallback (try `/workspaces/<nameOrId>` first and then `/<nameOrId>/workspaces/<nameOrId>` second if the first is a 403) as a minimal change, but this is a bit hamstrung by being annoying to test (you need to set up RBAC such that the fallback occurs) and because the internal `exists()` function obfuscates the status code. Given the problems there, I think we want to add a separate function to the workspace service with clarified docstrings, so that we end up with:

```
// Exists checks the existence of the Workspace name or ID in Kong. If RBAC is enabled, it requires permission to access the default workspace
func (s *WorkspaceService) Exists(ctx context.Context, nameOrID *string) (bool, error) {
... unchanged
}

// ExistsByName checks the existence of the Workspace name in Kong. If RBAC is enabled, it requires permission to access the <name> workspace
func (s *WorkspaceService) ExistsByName(ctx context.Context, name *string) (bool, error) {
... uses the "/<name>/workspaces/<name>"
}
```
And then in `workspace_service_test.go:TestWorkspaceService()`
```
existsName, err := client.Workspaces.ExistsByName(defaultCtx, createdWorkspace.name)
assert.Nil(err)
assert.True(exists)
```
Finally, KIC would use ExistsByName instead of Exists.

ccfishk:
> Ah, okay, my solution doesn't quite work with what this allows. You can only use the workspace name as the prefix, so this doesn't work if you try to use an ID, which is what the test checks:
> 
> ```
> > github.com/kong/go-kong/kong.(*WorkspaceService).Exists() ./workspace_service.go:53 (PC: 0x7f8225)
>     48:		if isEmptyString(nameOrID) {
>     49:			return false, fmt.Errorf("nameOrID cannot be nil for Get operation")
>     50:		}
>     51:	
>     52:		endpoint := fmt.Sprintf("/%v/workspaces/%v", *nameOrID, *nameOrID)
> =>  53:		return s.client.exists(ctx, endpoint)
>     54:	}
>     55:	
>     56:	// Create creates a Workspace in Kong.
>     57:	func (s *WorkspaceService) Create(ctx context.Context,
>     58:		workspace *Workspace) (*Workspace, error) {
> (dlv) p endpoint
> "/95b0679f-467c-4374-a161-7e639cba7e91/workspaces/95b0679f-467c-4374-a161-7e639cba7e91"
> ```
> 
> ```
> $ http localhost:8001/95b0679f-467c-4374-a161-7e639cba7e91/workspaces/95b0679f-467c-4374-a161-7e639cba7e91 
> HTTP/1.1 404 Not Found
> $ http localhost:8001/teamA/workspaces/95b0679f-467c-4374-a161-7e639cba7e91
> HTTP/1.1 200 OK
> ```
> 
> We can't differentiate names from IDs without fuzzy logic (strings that look parse as UUIDs are _probably_ IDs, but it's not guaranteed) or a breaking signature change to explicitly indicate what field you're requesting, which we don't want.
> 
> Alternatively this could use a fallback (try `/workspaces/<nameOrId>` first and then `/<nameOrId>/workspaces/<nameOrId>` second if the first is a 403) as a minimal change, but this is a bit hamstrung by being annoying to test (you need to set up RBAC such that the fallback occurs) and because the internal `exists()` function obfuscates the status code. Given the problems there, I think we want to add a separate function to the workspace service with clarified docstrings, so that we end up with:
> 
> ```
> // Exists checks the existence of the Workspace name or ID in Kong. If RBAC is enabled, it requires permission to access the default workspace
> func (s *WorkspaceService) Exists(ctx context.Context, nameOrID *string) (bool, error) {
> ... unchanged
> }
> 
> // ExistsByName checks the existence of the Workspace name in Kong. If RBAC is enabled, it requires permission to access the <name> workspace
> func (s *WorkspaceService) ExistsByName(ctx context.Context, name *string) (bool, error) {
> ... uses the "/<name>/workspaces/<name>"
> }
> ```
> 
> And then in `workspace_service_test.go:TestWorkspaceService()`
> 
> ```
> existsName, err := client.Workspaces.ExistsByName(defaultCtx, createdWorkspace.name)
> assert.Nil(err)
> assert.True(exists)
> ```
> 
> Finally, KIC would use ExistsByName instead of Exists.
{
Kong does not recognize workspace endpoints /ws-id/kong, /ws-id/rbac/users, it throws 
{"message":"Workspace 'abb80797-f338-4237-854e-7ea9dab7067a' not found"}. The ID only works under /workspaces/ws-id endpoints with admin permission

PR #93, opened 2021-09-30
feat: add key auth for admin api

geange:


rainest:
To handle this, instead of adding dedicated functionality here, we'd recommend just adding headers to the HTTP client you provide when creating the Kong client. We do something similar in the controller, where we provide a headers argument and inject those headers when building an HTTP client for our Kong client:

https://github.com/Kong/kubernetes-ingress-controller/blob/ba67e3e2690498f5ea8de16c9055cfeca9d5fa38/internal/manager/config.go#L106
https://github.com/Kong/kubernetes-ingress-controller/blob/ba67e3e2690498f5ea8de16c9055cfeca9d5fa38/internal/manager/config.go#L192
https://github.com/Kong/kubernetes-ingress-controller/blob/ba67e3e2690498f5ea8de16c9055cfeca9d5fa38/internal/adminapi/client.go#L68-L73

In addition to query string parameters, key-auth will look for any header with the same name as the key (as long as `key_in_header` is not disabled) and read its value as the key, which will work well with the standard client headers feature and avoids including keys in access logs. Those headers do still show up in Kong logging plugins, however, so I'd actually further recommend moving to basic-auth or some other plugin that uses the standard `Authorization` header always--our logging plugins recognize that and redact it, but won't recognize key-auth header names.

PR #97, opened 2021-10-29
feat(plugin) separate validation errors and fails

rainest:
Add a new string return value to PluginService.Validate(). When the
response is a validation failure, this string is the response body
message and the error is nil. Actual errors (e.g. connection timeouts,
response parse failures, etc.) still set the error.

Previously, failures also set the error. Downstream applications were
unable to determine whether the validation attempt was unsuccessful or
if the validation attempt was successful and confirmed the payload was
invalid.

This is a breaking API change for anything that uses Validate(). Not sure if that's anything other than the controller yet.

### DIFF
--- a/kong/plugin_service.go
+++ b/kong/plugin_service.go
@@ -28,7 +28,7 @@ type AbstractPluginService interface {
 	// ListAllForRoute fetches all Plugins in Kong enabled for a service.
 	ListAllForRoute(ctx context.Context, routeID *string) ([]*Plugin, error)
 	// Validate validates a Plugin against its schema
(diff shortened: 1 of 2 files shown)


shaneutt:
The change LGTM :+1: 

>This is a breaking API change for anything that uses Validate(). Not sure if that's anything other than the controller yet.

We're still in `v0` and as long as we bump to 23 I think that's all well and good, however if you're really interested there's [25 "used by" for this repo](https://github.com/Kong/go-kong/network/dependents?package_id=UGFja2FnZS0yMjc0Njg3NTIy) which we could consult.

PR #114, opened 2022-01-18
feat: add GetFullSchema and FillDefaults methods to plugins

GGabriele:
`GetFullSchema`: returns full schema for plugins (`/schemas/plugins/<ID>`)
`FillDefaults`: given a schema and a config object, fills all defaults

Based on the discussion happening in [here](https://github.com/Kong/deck/pull/562), this PR is introducing 2 new methods for `Plugin` (along few other helper functions, mostly taken from [KIC](https://github.com/Kong/kubernetes-ingress-controller/blob/main/internal/deckgen/deckgen.go#L121-L122) plus few changes).
These can be leveraged by `deck` and `KIC` to fill defaults values into plugins from their schemas.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -56,6 +56,8 @@ github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
+github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-querystring v1.1.0 h1:AnCroh3fv4ZBgVIf1Iwtovgjaw/GiKJo8M8yD/fhyJ8=
 github.com/google/go-querystring v1.1.0/go.mod h1:Kcdr2DB4koayq7X8pmAG4sNG59So17icRSOU623lUBU=
 github.com/google/gofuzz v1.0.0/go.mod h1:dBl0BpW6vV/+mYPU4Po3pmUjxk6FQPldtuIdl/M65Eg=
@@ -116,6 +118,12 @@ github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5
 github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 github.com/stretchr/testify v1.7.0 h1:nwc3DEeHmmLAfoZucVR881uASk0Mfjw8xYJ99tb5CcY=
 github.com/stretchr/testify v1.7.0/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
+github.com/tidwall/gjson v1.13.0 h1:3TFY9yxOQShrvmjdM76K+jc66zJeT6D3/VFFYCGQf7M=
+github.com/tidwall/gjson v1.13.0/go.mod h1:/wbyibRr2FHMks5tjHJ5F8dMZh3AcwJEMf5vlfC0lxk=
+github.com/tidwall/match v1.1.1 h1:+Ho715JplO36QYgwN9PGYNhgZvoUSc9X2c80KVTi+GA=
+github.com/tidwall/match v1.1.1/go.mod h1:eRSPERbgtNPcGhD8UCthc6PmLEQXEWd3PRB5JTxsfmM=
+github.com/tidwall/pretty v1.2.0 h1:RWIZEg2iJ8/g6fDDYzMpobmaoGh5OLl4AXtGUGPcqCs=
+github.com/tidwall/pretty v1.2.0/go.mod h1:ITEVvHYasfjBbM0u2Pg8T2nJnzm8xPwvNhhsoaGGjNU=
 github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.3.5/go.mod h1:mwnBkeHKe2W/ZEtQ+71ViKU8L12m81fl3OWwC1Zlc8k=
--- a/go.sum
+++ b/go.sum
@@ -55,8 +55,8 @@ github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMyw
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
-github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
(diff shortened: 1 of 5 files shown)


kong/plugin_service.go
```diff
@@ -29,14 +31,58 @@ type AbstractPluginService interface {
...
+	if plugin.Protocols == nil {
+		plugin.Protocols = getDefaultProtocols(gjsonSchema)
+	}
+	if plugin.Enabled == nil {
+		plugin.Enabled = Bool(true)
+	}
+	return plugin, nil
+}
+
+// GetSchema retrieves the config schema of a plugin
```
hbagdi, line 85: Mark this method as deprecated

kong/plugin_service.go
```diff
@@ -29,14 +31,58 @@ type AbstractPluginService interface {
...
+	// FillDefaults ingests plugin's defaults from its schema
+	FillDefaults(ctx context.Context, plugin *Plugin, schema map[string]interface{}) (*Plugin, error)
 }
 
 // PluginService handles Plugins in Kong.
 type PluginService service
 
-// GetSchema retrieves the schema of a plugin
+// GetFullSchema retrieves the full schema of a plugin
+func (s *PluginService) GetFullSchema(ctx context.Context,
```
hbagdi, line 46:
```suggestion
func (s *PluginService) GetPluginSchema(ctx context.Context,
```

kong/plugin_service.go
```diff
@@ -29,14 +31,58 @@ type AbstractPluginService interface {
...
+	var schema map[string]interface{}
+	_, err = s.client.Do(ctx, req, &schema)
+	if err != nil {
+		return nil, err
+	}
+	return schema, nil
+}
+
+// FillDefaults ingests plugin's defaults from its schema
+func (s *PluginService) FillDefaults(ctx context.Context,
```
hbagdi, line 65: This doesn't actually need to be on the PluginService since this function doesn't interact with the Admin API.

kong/plugin_service.go
rainest: IMO this should remain as-is: it's more descriptive of the difference between this and GetSchema, and we already know this is plugin stuff since it's part of the plugin service.

kong/plugin_service.go
rainest: Is the underlying endpoint going away? It's a bit confusing that the gateway has two similar, but slightly different endpoints for this, but if they're going to keep both I suppose we should as well, unless the richer endpoint returns a response that's compatible with any existing usage of the config-only endpoint, in which case we should just keep a single function and switch it to the richer endpoint.

rainest (commented):
What were the differences between this and the original KIC implementation? It looks like the main thing is that you get back a new plugin instance with the default config (which is fine), rather than getting a config and shoving that into your existing plugin, but I may be missing something since a lot of the code is similar.

kong/plugin_service.go
hbagdi: The underlying endpoint was supposed to go away when the new one was introduced.
It couldn't happen because a lot of reliance was on the old endpoint - something that we should change by doing what is suggested here.

> the richer endpoint returns a response that's compatible with any existing usage of the config-only endpoint

The new richer output although has what the old endpoint had to offer, it is different. For this reason, it would be best to have a new function and deprecate the old one. Users of this library then understand a lot more and can take appropriate precautions.

GGabriele:
The main difference with the KIC implementation is that KIC was filling only `Config` defaults, while here we are filling both `Config` and `Protocols`, as well as setting `Enabled` to `true` as default. This is done by leveraging the "newer" `/schemas/plugins/$ID` endpoint, so most of the code differences in the `fillConfigRecord` function are due to handling a slightly different output from the one used in KIC.

kong/plugin_service.go
GGabriele: Okay, makes sense to me, marked as deprecated

kong/plugin_service.go
GGabriele: Moved this under `kong/utils.go`

rainest (approved):

kong/plugin_service.go
```diff
@@ -29,14 +29,39 @@ type AbstractPluginService interface {
 	ListAllForRoute(ctx context.Context, routeID *string) ([]*Plugin, error)
 	// Validate validates a Plugin against its schema
 	Validate(ctx context.Context, plugin *Plugin) (bool, string, error)
-	// GetSchema retrieves the schema of a plugin
+	// GetSchema retrieves the config schema of a plugin
```
hbagdi, line 32:
```suggestion
	// GetSchema retrieves the config schema of a plugin.
```

kong/plugin_service.go
```diff
@@ -29,14 +29,39 @@ type AbstractPluginService interface {
...
+	//
+	// Deprecated: Use GetFullSchema instead
```
hbagdi, line 34:
```suggestion
	// Deprecated: Use GetFullSchema instead.
```

kong/plugin_service.go
```diff
@@ -29,14 +29,39 @@ type AbstractPluginService interface {
...
 	GetSchema(ctx context.Context, pluginName *string) (map[string]interface{}, error)
+	// GetFullSchema retrieves the full schema of a plugin
```
hbagdi, line 36:
```suggestion
	// GetFullSchema retrieves the full schema of a plugin.
	// This makes the use of `/schemas` endpoint in Kong.
```

kong/plugin_service.go
```diff
@@ -29,14 +29,39 @@ type AbstractPluginService interface {
...
+	GetFullSchema(ctx context.Context, pluginName *string) (map[string]interface{}, error)
 }
 
 // PluginService handles Plugins in Kong.
 type PluginService service
 
-// GetSchema retrieves the schema of a plugin
+// GetFullSchema retrieves the full schema of a plugin
```
hbagdi, line 43:
```suggestion
// GetFullSchema retrieves the full schema of a plugin.
```

kong/plugin_service_test.go
```diff
@@ -320,6 +321,118 @@ func TestPluginListAllForEntityEndpoint(T *testing.T) {
...
+	assert.Nil(err)
+	assert.NotNil(client)
+
+	tests := []struct {
+		name     string
+		plugin   *Plugin
+		expected *Plugin
+	}{
+		{
+			name: "no_config_no_protocols",
```
hbagdi, line 337: Remove all `_`.

go.sum
```diff
@@ -56,6 +56,8 @@ github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
 github.com/google/go-cmp v0.5.5 h1:Khx7svrCpmxxtHBq5j2mp/xVjsi8hQMfNLvJFAlrGgU=
 github.com/google/go-cmp v0.5.5/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/google/go-cmp v0.5.6 h1:BKbKCqvP6I+rmFHt06ZmyQtvB8xAkWdhFyr0ZUNZcxQ=
+github.com/google/go-cmp v0.5.6/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
```
hbagdi, line 60: Run `go mod tidy`

kong/utils.go
```diff
@@ -141,3 +143,91 @@ func VersionFromInfo(info map[string]interface{}) string {
...
+
+func getDefaultProtocols(schema gjson.Result) []*string {
+	var res []*string
+	fields := schema.Get("fields")
+
+	for _, field := range fields.Array() {
+		for key, value := range field.Map() {
+			if key != "protocols" {
+				continue
+			}
```
hbagdi, line 155: Map() returns a map[string]Result:
https://pkg.go.dev/github.com/tidwall/gjson?utm_source=godoc#Result.Map

Please remove this loop and instead use `field.Map()["protocols"]` to do the same thing.
Linear searches are best avoided, especially when an index is already available.

hbagdi (changes requested):
Add an integration test for the GetFulLSchema function that has been added. The test must ensure that this function works against OSS as well as Enterprise.

GGabriele:
> Add an integration test for the GetFulLSchema function that has been added. The test must ensure that this function works against OSS as well as Enterprise.

isn't it already tested on both as part of this [here](https://github.com/Kong/go-kong/pull/114/files#diff-d2200bbc92069856fbd979a168f1a70b8ccdbbadbd23fe3998f94afcf8779868R422)?

hbagdi:
> isn't it already tested on both as part of this here?

Oh boy.
That test is for another function. If that function is removed, we would be removing the call to the GetFullSchema function.

Please add another test that only tests the GetFullSchema method.
Assert the response code, and check for a few elements in the body to make sure the response is as expected.
Also, add a negative test for the "noexists" plugin.

GGabriele:
Wanted to add a testcase to check the whole schema response, but couldn't find a plugin whose schema remained fully consistent across the various EE-OSS versions, so I reverted to just test errors and basic response structure. Let me know if that's fine.

hbagdi (approved):

kong/utils.go
```diff
@@ -141,3 +143,87 @@ func VersionFromInfo(info map[string]interface{}) string {
...
+			res[fname] = nil
+		}
+		return true
+	})
+
+	return res
+}
+
+// FillPluginsDefaults ingests plugin's defaults from its schema
+func FillPluginsDefaults(plugin *Plugin, schema map[string]interface{}) (*Plugin, error) {
```
hbagdi, line 212: @GGabriele I made a mistake in the review.
This method interface is misleading and wrong.
The code takes in a plugin struct and then mutates it.
But the method signature also returns a plugin resource.

Either of the following are fine (but together not):
- Return only an error, and document that the plugin is filled and mutated
- Inside the function, make a copy of the input plugin, mutate the copy and return it

GGabriele: Doh, my bad, sorry :(

https://github.com/Kong/go-kong/pull/116

PR #118, opened 2022-01-24
tests: add new ee versions to test matrix

GGabriele:


rainest (changes requested):
We only care about having a single patch version of any given release, so instead of

```
        - 'enterprise-2.3.3.3'
        - 'enterprise-2.3.3.4'
```
we only want to include `- 'enterprise-2.3.3.4'`, and so on.

GGabriele:
> We only care about having a single patch version of any given release, so instead of
> 
> ```
>         - 'enterprise-2.3.3.3'
>         - 'enterprise-2.3.3.4'
> ```
> 
> we only want to include `- 'enterprise-2.3.3.4'`, and so on.

Makes sense, thanks for taking a look!

Looks better now?

hbagdi (approved):

PR #119, opened 2022-01-24
feat: add SchemaService and FillEntityDefaults utility

GGabriele:
Mirroring what's been previously done for `Plugins`,
this adds `GetFullSchema` to `Routes`, `Upstreams`, `Targets` and `Services`.
This also adds utils function to fill core entities with
defaults comings from their schema.


hbagdi:
Let's do this a bit differently.
Let's create a SchemaService with a function to pull a schema for any entity. `Get(entityname string)` something like. That is much more concise and meets the goal.

GGabriele:
@hbagdi good point.

Since I removed the `GetFullSchema` from each entity service, I also moved all the tests under `kong/utils_test.go`. Let me know what you think.

kong/schema_service.go
```diff
@@ -0,0 +1,29 @@
...
+type AbstractSchemaService interface {
+	// Get fetches an entity schema from Kong.
+	Get(ctx context.Context, entity string) (map[string]interface{}, error)
+}
+
+// SchemaService handles schemas in Kong.
+type SchemaService service
+
+// Get retrieves the full schema of kong entities.
+func (s *SchemaService) Get(ctx context.Context, entity string) (map[string]interface{}, error) {
```
hbagdi, line 18: make a type 
```
type map[string]interface{} Schema
```
return that here and then take this as input to the util package. Might result in import cycles.

kong/schema_service_test.go
```diff
@@ -0,0 +1,23 @@
+package kong
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGet(T *testing.T) {
```
hbagdi, line 9: this should follow the naming convention found in other service tests.

kong/schema_service_test.go
```diff
@@ -0,0 +1,23 @@
...
+	assert := assert.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.Nil(err)
+	assert.NotNil(client)
+
+	entities := []string{"services", "routes", "targets", "upstreams", "plugins"}
```
hbagdi, line 16: add all core entities, consult documentation or kong/db/schemas/entities dir in kong/kong repo

kong/utils.go
```diff
@@ -208,6 +214,167 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	defaults := flattenDefaultsSchema(gjsonSchema)
+	jsonSchemaWithDefaults, err := json.Marshal(&defaults)
+	if err != nil {
+		return nil, err
+	}
+	return jsonSchemaWithDefaults, nil
+}
+
+// FillUpstreamsDefaults ingests upstreams' defaults from their schema.
+func FillUpstreamsDefaults(upstream *Upstream, schema map[string]interface{}) error {
```
hbagdi, line 299: Since the code doesn't actually deal with the type of the first argument, can we have a generic function that takes in an "interface{}" and does the right thing based on type of the input.
You will need to use a "type switch" in Go.

kong/schema_service_test.go
GGabriele: Done, although I had to skip some entities, as `clustering_data_planes` and `parameters` are not present in < `2.3` and `workspaces` are only EE.

We should probably have a way to conditionally include or exclude test cases based on the running Kong version.

kong/utils.go
GGabriele: 🧙‍♂️

kong/schema_service_test.go
hbagdi: > Done, although I had to skip some entities, as clustering_data_planes and parameters are not present in < 2.3 and workspaces are only EE.

No problem.

> We should probably have a way to conditionally include or exclude test cases based on the running Kong version.

We do. See runWhen* funcs.

kong/utils.go
```diff
@@ -208,9 +214,113 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	if err = json.Unmarshal(defaults, &tmpEntity); err != nil {
+		return err
+	}
+	return mergo.Merge(entity, tmpEntity)
+}
+
 // FillPluginsDefaults ingests plugin's defaults from its schema.
 // Takes in a plugin struct and mutate it in place.
-func FillPluginsDefaults(plugin *Plugin, schema map[string]interface{}) error {
+func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
```
hbagdi, line 323: We don't return Schema from plugin.GetFullSchema. Can we change that too?

kong/utils.go
```diff
@@ -208,9 +214,113 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	case *Target:
+		tmpEntity = &Target{}
+	case *Service:
+		tmpEntity = &Service{}
+	case *Route:
+		tmpEntity = &Route{}
+	case *Upstream:
+		tmpEntity = &Upstream{}
+	}
+	defaults, err := getDefaultsObj(schema)
```
hbagdi, line 311: check for schema == nil before using

kong/utils.go
```diff
@@ -208,9 +214,113 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	if err != nil {
+		return err
```
hbagdi, line 313: Annotate error,
```suggestion
		return fmt.Errorf("parse schema for defaults: %v", err)
```

kong/utils.go
```diff
@@ -208,9 +214,113 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	}
+	if err = json.Unmarshal(defaults, &tmpEntity); err != nil {
+		return err
```
hbagdi, line 316: Annotate error.

kong/utils.go
```diff
@@ -208,9 +214,113 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	}
+	return mergo.Merge(entity, tmpEntity)
```
hbagdi, line 318: Annotate error

kong/utils.go
```diff
@@ -208,9 +214,113 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+	jsonSchema, err := json.Marshal(&schema)
+	if err != nil {
+		return nil, err
+	}
+	gjsonSchema := gjson.ParseBytes((jsonSchema))
+	defaults := flattenDefaultsSchema(gjsonSchema)
+	jsonSchemaWithDefaults, err := json.Marshal(&defaults)
+	if err != nil {
+		return nil, err
+	}
```
hbagdi, line 294: Code does a JSON marshal and unmarshal of defaults to get it into the typed structs. This can be fairly expensive, especially in hot code paths. The alternative is to use reflections.
No change is needed in the patch, only commenting to keep this in mind for the future.

kong/utils_test.go
```diff
@@ -159,3 +161,356 @@ func Test_requestWithHeaders(t *testing.T) {
...
+	assert.Nil(err)
+	assert.NotNil(client)
+
+	tests := []struct {
+		name     string
+		route    *Route
+		expected *Route
+	}{
+		{
+			name: "name paths",
```
hbagdi, line 178:
```suggestion
			name: "fills defaults for all fields except paths, leaves name as is",
```
Descriptive tests, please.

hbagdi (commented):
Nice patch!

kong/utils.go
GGabriele: Mmmh what do you think it's best here if `schema == nil`, returning `nil` or erroring out complaining about it being `nil`? Maybe the former?

kong/utils.go
GGabriele: sure

kong/utils.go
hbagdi: Return an error. When in doubt, return an error to the caller. That leads to better UX because the caller can then appropriate action. If the code returns without an error, the caller may assume that the function call worked only to be surprised later that it didn't.

kong/utils_test.go
```diff
@@ -159,3 +161,356 @@ func Test_requestWithHeaders(t *testing.T) {
...
+				Paths:                   []*string{String("/r1")},
+				PreserveHost:            Bool(false),
+				Protocols:               []*string{String("grpc")},
+				RegexPriority:           Int(0),
+				StripPath:               Bool(true),
+				HTTPSRedirectStatusCode: Int(426),
+			},
+		},
+		// TODO: re-add once upstream patch has been merged.
+		// https://github.com/imdario/mergo/pull/203
```
hbagdi, line 211: It is likely that we will forget about this TODO. Can you remove the code and open a Github issue to track it?

kong/utils.go
GGabriele: roger that

kong/utils_test.go
GGabriele: https://github.com/Kong/go-kong/issues/122

hbagdi:
@rainest Can I get a review from you on this one as well?

hbagdi (approved):
Let's move ahead on this one. If there are more issues, @GGabriele please take care of resolving with subsequent PRs.

PR #120, opened 2022-01-25
feat: add missing entities fields

GGabriele:


hbagdi:
@rainest Can you review this with some git spelunking? Scope includes all relevant entities until Kong 2.8.

rainest:
Do you mean specific versions? Something like `git diff 2.5.0 -- kong/db/schema/entities/` can tell you what changed between versions, but there's no way to easily know if there's an entity in go-kong missing a field from a version.

hbagdi:
I meant that could you please do that "git diff" and review this PR to ensure that go-kong is not missing anything that was added in Kong 2.x series (or at least not in the last few releases)?

rainest:
Looking back to 2.0 (since cert digest is quite old, added in 2.1), manual review did not find anything other than service.enabled, cert digest, and cert/key alts that we're missing in supported core types.

There are new unsupported types:
- the clustering_data_plane entity, which is only used for reporting status
- the parameter entity, which looks like an internal helper you don't interact with directly.

All other changes are constraint changes or minor type changes that would not affect the JSON types.

rainest (approved):

hbagdi (approved):

hbagdi:
Looking forward to a world where this wouldn't be required! ;)

PR #148, opened 2022-03-17
fix: marshall and unmarshall actions on endpoint permissions

GGabriele:
For endpoint permissions's `actions`, Kong returns a list with GET requests, but it expects a
comma-separated string with POST/PUT requests.

This fixes the marshalling for this entity (reverting https://github.com/Kong/go-kong/pull/131)

Sample script attempting to configure `actions` as a list:

```
$ echo '
id: 82bfb84b-d659-4ac1-ab29-94576f838753
name: workspace-super-admin
comment: "Full access to all endpoints in the workspace"
' | y2j |  http PUT ":8001/workspace1/rbac/roles/82bfb84b-d659-4ac1-ab29-94576f838753"

$ echo '
workspace: workspace1
endpoint: /rbac/*/*/*
actions:
    - delete
    - create
    - read
    - update
negative: true
' | y2j | http POST ":8001/workspace1/rbac/roles/82bfb84b-d659-4ac1-ab29-94576f838753/endpoints"

HTTP/1.1 200 OK
{
    "comment": "Full access to all endpoints in the workspace",
    "created_at": 1647520491,
    "id": "82bfb84b-d659-4ac1-ab29-94576f838753",
    "is_default": false,
    "name": "workspace-super-admin"
}


HTTP/1.1 201 Created
{
    "actions": [],
    "created_at": 1647520491,
    "endpoint": "/rbac/*/*/*",
    "negative": true,
    "role": {
        "id": "82bfb84b-d659-4ac1-ab29-94576f838753"
    },
    "workspace": "workspace1"
}
```

The `actions` field is ignored. While it works if the `actions` field is a string:

```
$ echo '
workspace: workspace1
endpoint: /rbac/*/*/*
actions: delete,create,read,update
negative: true
' | y2j | http POST ":8001/workspace1/rbac/roles/82bfb84b-d659-4ac1-ab29-94576f838753/endpoints"
HTTP/1.1 201 Created
{
    "actions": [
        "read",
        "delete",
        "create",
        "update"
    ],
    "created_at": 1647520694,
    "endpoint": "/rbac/*/*/*",
    "negative": true,
    "role": {
        "id": "82bfb84b-d659-4ac1-ab29-94576f838753"
    },
    "workspace": "workspace1"
}
```

### DIFF
--- a/kong/types.go
+++ b/kong/types.go
@@ -3,6 +3,7 @@ package kong
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 )
 
 // Service represents a Service in Kong.
@@ -363,17 +364,20 @@ func (e *RBACEndpointPermission) MarshalJSON() ([]byte, error) {
 		CreatedAt *int      `json:"created_at,omitempty" yaml:"created_at,omitempty"`
 		Workspace *string   `json:"workspace,omitempty" yaml:"workspace,omitempty"`
 		Endpoint  *string   `json:"endpoint,omitempty" yaml:"endpoint,omitempty"`
-		Actions   []*string `json:"actions,omitempty" yaml:"actions,omitempty"`
+		Actions   *string   `json:"actions,omitempty" yaml:"actions,omitempty"`
 		Negative  *bool     `json:"negative,omitempty" yaml:"negative,omitempty"`
 		Role      *RBACRole `json:"role,omitempty" yaml:"role,omitempty"`
 		Comment   *string   `json:"comment,omitempty" yaml:"comment,omitempty"`
 	}
-
+	var actions []string
+	for _, action := range e.Actions {
+		actions = append(actions, *action)
+	}
 	return json.Marshal(&ep{
 		CreatedAt: e.CreatedAt,
 		Workspace: e.Workspace,
 		Endpoint:  e.Endpoint,
-		Actions:   e.Actions,
+		Actions:   String(strings.Join(actions, ",")),
 		Negative:  e.Negative,
 		Comment:   e.Comment,
 	})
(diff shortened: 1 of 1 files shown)


shaneutt (commented):
Looks good overall, but given the context that we're fixing something we broke before it seems like we should be updating or adding tests which will protect this against future regressions. What are your thoughts?

GGabriele:
Totally! Not sure why, but it seems that we are skipping tests on RBAC [here](https://github.com/Kong/go-kong/blob/main/kong/test_utils.go#L67-L70).

Since now we have a "proper" testing pipeline in `decK`, what I plan to do is to keep adding more and more test cases to give us a good coverage. Like, [this](https://github.com/Kong/deck/pull/619/commits/de48b7c8652a3e6b81b7c807837f3eaab158fc46).

rainest:
It skips them if RBAC is not available, which we expect for community. It _should_ be on for Enterprise, but for some reason https://github.com/Kong/go-kong/blob/bd63e507f97052392a391f91766c766fcaf465c3/.ci/setup_kong.sh#L63-L66 is apparently having no effect, despite not being under any conditional.

Dunno when/why that stopped working, but I suppose there's no reason to not just set them in the Action environment instead; they're all static anyway... Except that doesn't appear to work either.

If it did, https://github.com/Kong/go-kong/blob/bd63e507f97052392a391f91766c766fcaf465c3/kong/rbac_user_service_test.go#L120-L123 would test at least part of this.

rainest:
IDK _why_ those variables are getting ignored (they do show up in `env` output run before the Kong commands), but setting them on the invocation line instead of exporting them works, so whatever.

However fixing this exposes that there were several (mostly) unrelated broken tests lurking around: https://github.com/Kong/go-kong/runs/5636336474?check_suite_focus=true

GGabriele:
Also, those variable were set even before your commit: https://github.com/Kong/go-kong/blob/main/.ci/setup_kong.sh#L65-L66

How do we want to proceed here? Add these vars directly in the invocation and fix those failing tests?

rainest:
> How do we want to proceed here? Add these vars directly in the invocation and fix those failing tests?

Yeah--I don't really much care what the setup script looks like so long as it works, and inline does. I'm curious why that's the case (sudo would be my first guess, but sudo does inherit the environment of its caller by default, which a run of `env` confirms, so not sure what's up there), but not to the extent that I'll dig into it given a working alternative.

With that and a bunch of new changes unrelated to the original PR, tests pass, add (and actually run) a regression test for this, and fixes any tests that were broken.

https://github.com/Kong/go-kong/pull/148/commits/430956c23d03c377f712fed48d19b742bce018bc tests the original fix condition, as I understand it.

The rest of the changes fix broken tests that were hidden by Enterprise tests not actually running for the most part. Other than https://github.com/Kong/go-kong/pull/148/commits/c35a1dd41aabe9a6d4b120f1ad1a1a4efd503830, they do not change library code, only test code.

Most fixes are unremarkable; https://github.com/Kong/go-kong/pull/148/commits/6ae5da8cfd6fc3a70782906bbb3919f4fe875ce7 is the exception. The implementation of admin consumers and credentials is weird, and apparently changes the way those resources paginate even though the created resources don't exist from the perspective of the admin API. We had [acknowledged as much before](https://github.com/Kong/go-kong/blame/bd63e507f97052392a391f91766c766fcaf465c3/kong/basic_auth_service_test.go#L303-L306) but apparently didn't actually fix it there and didn't notice because of the skips.

Checking for specific pagination behavior isn't something anyone should need to do in practice--all you should care about as a library user is that you can pass in a next value and get the expected page. We should consider reworking our tests to not require specific pagination behavior, but IMO it's fairly low-value: this endpoint exists in community doesn't change its behavior on Enterprise, so there's no pressing reason to test it on Enterprise. The specific pages change depending on your consumer set, but that's expected--right now our test requires a specific consumer set, and it arguably shouldn't.

PR #150, opened 2022-03-21
feat(client) add configuration_hash to Status

rainest:
Adds support for the configuration hash added to 2.8's status endpoint.


shaneutt (approved):
I would just say make sure we do a manual check against an image that has the hash before merging and then :+1:

rainest:
Waiting on https://github.com/Kong/kubernetes-ingress-controller/actions/runs/2018716499, which will test this by virtue of requiring this go-kong feature to work.

rainest:
Right then, overall test failed, but because some unrelated test had a bug uncovered by a fix elsewhere. The test that matters for this succeeded.

PR #152, opened 2022-03-22
fix: handle auto fields in plugins schema

GGabriele:
This was highlighted in an [issue](https://github.com/Kong/deck/issues/605) raised in decK.

Sample config:

```
$ cat kong.yaml
plugins:
- name: rate-limiting-advanced
  config:
    limit:
    - 3
    sync_rate: 10
    window_size:
    - 60
```

`rate-limiting-advanced` has a `namespace` config field which is set as `auto`:

```
$ http :8001/schemas/plugins/rate-limiting-advanced
...
...
                        "namespace": {
                            "auto": true,
                            "required": true,
                            "type": "string"
                        }
                    },
...
...
```

Since no `default` entry is found in the schema, the field is left empty, which causes a problem since the field is also marked as `required`.

```
$ deck sync
creating plugin rate-limiting-advanced (global)
Summary:
  Created: 0
  Updated: 0
  Deleted: 0
Error: 1 errors occurred:
	while processing event: {Create} plugin rate-limiting-advanced (global) failed: HTTP status 400 (message: "schema violation (config.namespace: required field missing)")
```

This PR makes sure that we auto-generates a value for such fields as well if not provided by the user

```
$ ./deck sync
creating plugin rate-limiting-advanced (global)
Summary:
  Created: 1
  Updated: 0
  Deleted: 0
```

Question: do all `auto` fields expects a 32 chars string? The answer was not obvious to me while navigating Kong codebase

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,5 +13,6 @@ require (
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/stretchr/testify v1.7.1
 	github.com/tidwall/gjson v1.14.0
+	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/code-generator v0.23.5
 )
--- a/kong/plugin_service_test.go
+++ b/kong/plugin_service_test.go
@@ -459,6 +459,45 @@ func TestFillPluginDefaults(T *testing.T) {
 	}
 }
 
+func TestFillPluginDefaultsWithAutoFields(T *testing.T) {
+	assert := assert.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.Nil(err)
+	assert.NotNil(client)
+
+	tests := []struct {
+		name      string
+		plugin    *Plugin
+		autoField string
+	}{
+		{
+			name: "oauth2 has 'provision_key' auto field",
+			plugin: &Plugin{
+				Name: String("oauth2"),
+				Config: Configuration{
+					"enable_authorization_code": true,
+				},
+			},
+			autoField: "provision_key",
+		},
+	}
+
+	for _, tc := range tests {
+		T.Run(tc.name, func(t *testing.T) {
+			p := tc.plugin
+			fullSchema, err := client.Plugins.GetFullSchema(defaultCtx, p.Name)
+			assert.Nil(err)
+			assert.NotNil(fullSchema)
+			if err := FillPluginsDefaults(p, fullSchema); err != nil {
+				t.Errorf(err.Error())
+			}
+			// auto field shouldn't be nil
+			assert.NotNil(p.Config[tc.autoField])
+		})
+	}
+}
+
 func comparePlugins(expected, actual []*Plugin) bool {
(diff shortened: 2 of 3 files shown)


hbagdi:
Auto-generating fields will render the desired state to be out of date, meaning, subsequent  `deck diff` will always detect a diff.
The right solution here is to ask the user to provide a value. The default injection in decK is for static defaults only.

rainest:
Since we know the field is auto from the schema, can we not instead treat any value as valid unless it's explicitly specified by (and does not match) the target state?

Philosophically, yes, this means that your configuration is not fully declarative, but that's partly the status quo AFAIK (since we'll try to delete it, Kong will regenerate it with some new value you didn't specify), and the status quo is worse than allowing any value (e.g. here if you leave the namespace alone, allowing any value lets the plugin function normally, whereas deleting it means that you reset all your counters every config update).

It's true that many auto fields may not make sense in deck workflows. Auth plugin secrets are the most common case for auto, and it seems strange that you'd create one with deck and then go retrieve its actual value from the admin API after. rate-limiting is a bit odd in that you'd normally only choose a specific namespace to reduce counter usage or to share counters between plugin instances. For the credential case, I'd expect users to go ahead and specify values anyway, but for this a random value is fine, since clients don't need to know it.

If we don't allow random values, we should reject configuration if an auto field is empty, and would need to replicate that in the ingress controller admission webhook. I don't think there's ever a good case for the current churn behavior.

hbagdi:
Given a schema with an auto field, decK or KIC can't know if the field should be diffed or ignored, unless that information is present in the schema or is hard-coded. Given the absence of such information, it makes most sense to not ignore auto fields and instead ask the user to provide a value for them.

rainest:
Do we lack sufficient information to make a decision? This seems clear enough to me:
- If a field is present in target state, always sync to match the target state.
- If a field has auto set in the schema, is absent from target state, and present in current config, do nothing/do not attempt a sync that clears that field.
- If a field does not have auto set in the schema, is absent from the target state, and present in the current config, sync to clear the field.

We have the info needed to make a decision, though it's probably a bit difficult to plumb the schema auto info down to where we'd need it for that. Rejecting empty auto fields is probably easier, since we should be able to handle that as an error case within the defaulter, where we're already operating on schema info prior to diffing.

hbagdi:
There is another reason to not do this: the schemas as we know of are going to go away in the future and the level of detail in the schema won't be available to decK.

GGabriele:
In summary, should we just leave things as they are and let Kong return an error to decK/KIC? Or should we refactor the `fillConfigRecord` a bit to return a custom error? The latter may be better as the `(message: "schema violation (config.namespace: required field missing)")` message from Kong doesn't tell the user why such field is not automatically filled.

hbagdi:
I personally think that the error returned from Kong has enough details. decK/KIC could point the user to the docs site.
Parsing schema more than necessary is best avoided at the moment.

PR #167, opened 2022-05-06
feat: support both POST and PUT for target entities

GGabriele:
Starting with 3.0, POSTs are not supported anymore for the `/targets`
endpoints and they will be replaced by PUTs. This means that we need
a way to support both methods in order to offer backward compatibility.

This introduces a 'retry' mechanism to be leveraged in the case of
POST returning 405s.

### DIFF
--- a/kong/target_service.go
+++ b/kong/target_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 )
 
 // AbstractTargetService handles Targets in Kong.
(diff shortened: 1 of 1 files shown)


rainest:
Any reason not to instead make a root endpoint call, run `VersionFromInfo()`, and store the version as part of the client during initialization, then using a `getTargetCreationMethod(version)` helper? Target creation happens quite often (at least in the KIC context), so we'd like to avoid doubling requests for those on modern nodes.

That does add a dependency on access to the root endpoint (we do not have one yet), which we can't actually guarantee in RBAC instances, so alternatively we could cache the result of a feature-based check, i.e. check if the correct method is known already and use it if so, else fall back to the system that tries and sees if it encounters a 405, and stores the correct method based on whether it does or not.

hbagdi:
Please don't merge this just yet. Talking to Gabriele out of band right now.

PR #172, opened 2022-06-01
ci: add Github Actions' cron to update Kong license

pmalek:
This adds a Github Actions' cron which will try to update the secret with Kong license (named: `KONG_LICENSE_DATA`) by pulling it from Pulp.

Apart from pulp related secrets (that contain credentials and the URL) this workflow requires [Github Personal Token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) with `repo` permissions.

### DIFF
--- a/.github/workflows/kong-license.yaml
+++ b/.github/workflows/kong-license.yaml
@@ -1,0 +1,29 @@
+name: Update Kong license secret
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch: {}
+
+jobs:
+  test:
+    env:
+      PULP_USERNAME: ${{ secrets.PULP_USERNAME}}
+      PULP_PASSWORD: ${{ secrets.PULP_PASSWORD}}
+      PULP_LICENSE_URL: ${{ secrets.PULP_LICENSE_URL}}
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: License
+        id: license
+        run: |
+          LICENSE=$(curl -s -L -u"$PULP_USERNAME:$PULP_PASSWORD" "$PULP_LICENSE_URL")
+          echo "::add-mask::$LICENSE"
+          echo "::set-output name=LICENSE::$LICENSE"
+
+      - uses: gliech/create-github-secret-action@ea87807ab20663b30a1a2d14d7f6dd9490b1e7a1 # v1.4.10
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/kong-license.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+    steps:
+      - name: License
+        id: license
+        run: |
+          LICENSE=$(curl -s -L -u"$PULP_USERNAME:$PULP_PASSWORD" "$PULP_LICENSE_URL")
+          echo "::add-mask::$LICENSE"
+          echo "::set-output name=LICENSE::$LICENSE"
+
+      - uses: gliech/create-github-secret-action@v1
```
shaneutt, line 25: I'm a little iffy about this dependency. I'm not saying no, but I'm curious how hard it would be to put a simple curl one liner to update the license instead of using this dep?

pmalek: It would better for sure but it seems you need to

- get the key used for secrets encryption
- use the key to encrypt the secret contents
- call [github API](https://docs.github.com/en/rest/actions/secrets#create-or-update-a-repository-secret) with said encrypted contents

It's not rocket science but not-trivial either.

ref: https://blog.knoldus.com/playing-with-github-secret-api/#2-get-an-organization-public-key

shaneutt: I gotcha. @rainest you got any thoughts on this one?

shaneutt: Sounds good, let's pin this and I would consider this comment resolved.

shaneutt (approved):

.github/workflows/kong-license.yaml
pmalek: Done, PTAL at the updated version.

.github/workflows/kong-license.yaml
```diff
@@ -0,0 +1,29 @@
+name: Update Kong license secret
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch: {}
+
+jobs:
+  test:
```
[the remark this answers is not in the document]
shaneutt: woops :shrug:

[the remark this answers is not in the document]
shaneutt: Just put in another PR nbd.

PR #173, opened 2022-06-02
chore(tests): prevent seg faults in tests by using proper assertions

pmalek:


rainest (approved):

rainest:
ACL test weirdness now properly accounted for in https://github.com/Kong/go-kong/issues/176

Whatever caused it, it isn't this. The assert changes are the same on Enterprise versions that do not hit that issue (<2.2). Going ahead and merging.

PR #183, opened 2022-06-22
feat: rate-limit requests when receiving 429s

GGabriele:


hbagdi:
Magically making the existing client retry on 429 would be a surprising behavior.
I'm also not sure if this is the desired behavior. Can we have the 429 be returned to the user and the user can retry if the need be?

This client has mostly been a barebones binding client without superpowers for the most parts, we should keep it that way.

PR #187, opened 2022-07-06
feat(plugins) add support for ordering

rainest:
Adds Go bindings for the [plugin ordering feature](https://github.com/Kong/kong-ee/pull/3164/files#diff-70d5027c92c8ca9908b7bdcaf015d79e2c70ba1713a19dff9fe9fcbf19d0c6fbR645-R669) (see https://github.com/Kong/docs.konghq.com/pull/3905 also). Requesting feedback on ideas for a succinct name and description of the nested type: trying to describe it independent of its parent has broken my brain.

My review of this feature for KIC is that it's just an additional field on plugin entities and shouldn't need anything more complex than adding these types directly to Kong(Cluster)Plugin. These do not include validation annotations because AFAIK we cannot add constraints for map keys and these technically aren't static--we rely on the Kong validation endpoint to tell us if a specific version supports overrides in phases other than `access` (unlike this initial version).

IMO The one design decision here is that `Before` and `After` are fields whereas the phases are not, and simply use a map. My rationale is that the order options are almost certainly static (what would you add beyond before and after?) whereas the _allowed_ phases may vary per version. The complete list of phases is (essentially) a static feature of lua-nginx-module, but it seemed silly to add a bunch of fields that you can't actually use (and may never be able to use) that are all the same type anyway.

go-kong is weird and does not use Docker for its test infrastructure, and instead installs packages directly into the test environment. We discard Enterprise deb packages after building nightly images per Colin H., so we won't actually be able to run this test in CI until 3.0 is released without overhauling this test suite. I recommend we rely on downstream tests (i.e. KIC) instead for the interim. To demonstrate that the test here is valid, here's a local run against a 3.0 nightly:

```
11:39:35-0700 esenin $ git diff                         
diff --git a/kong/plugin_service_test.go b/kong/plugin_service_test.go
index 4d2d65b..74c7812 100644
--- a/kong/plugin_service_test.go
+++ b/kong/plugin_service_test.go
@@ -109,6 +109,7 @@ func TestPluginWithTags(T *testing.T) {
 
 func TestPluginWithOrdering(T *testing.T) {
        RunWhenEnterprise(T, ">=3.0.0", RequiredFeatures{})
+       T.Log("alert! blue guys and brainsuckers have entered the base!")
        assert := assert.New(T)
 
11:39:59-0700 esenin $ http localhost:8001 | jq .version
"3.0.0-4825cd1e7-enterprise-edition"
11:40:03-0700 esenin $ go test ./... -run TestPluginWithOrdering -v
=== RUN   TestPluginWithOrdering
    plugin_service_test.go:112: alert! blue guys and brainsuckers have entered the base!
--- PASS: TestPluginWithOrdering (0.05s)
PASS
ok  	github.com/kong/go-kong/kong	0.049s
testing: warning: no tests to run
PASS
ok  	github.com/kong/go-kong/kong/custom	(cached) [no tests to run]
```


kong/plugin_service_test.go
```diff
@@ -107,6 +107,36 @@ func TestPluginWithTags(T *testing.T) {
...
+	assert.Equal(PluginOrdering{
+		Before: PluginOrderingPhase{
+			"access": []string{"key-auth", "basic-auth"},
+		},
+	}, *createdPlugin.Ordering)
+
+	err = client.Plugins.Delete(defaultCtx, createdPlugin.ID)
+	assert.NoError(err)
+}
+
```
jrsmroz, line 139: Trying to understand what this is testing. There is no new functionality added in this PR. Are testing marshalling?

kong/plugin_service_test.go
```diff
@@ -107,6 +107,36 @@ func TestPluginWithTags(T *testing.T) {
...
+	assert := assert.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
+
+	plugin := &Plugin{
+		Name: String("request-termination"),
+		Ordering: &PluginOrdering{
+			Before: PluginOrderingPhase{
```
GGabriele, line 121: Can we also add a test using `After` and one using both?

kong/plugin_service_test.go
```diff
@@ -107,6 +107,36 @@ func TestPluginWithTags(T *testing.T) {
...
+				"access": []string{"key-auth", "basic-auth"},
```
GGabriele, line 122: What do you think about testing that:
- adding a not-existing plugin in `PluginOrderingPhase` errors out
- adding a random key in `PluginOrderingPhase` errors out

kong/plugin_service_test.go
rainest: Essentially. Pretty much everything in go-kong is marshalling and unmarshalling 🙂

Our utilities for version-dependent tests affect the whole test, so this is separate to let it run on 3.0+ only, similar to the separate tag test.

jrsmroz: Thanks

kong/plugin_service_test.go
rainest: 9723ed5 just does a single one with both. For the successful case we just care that all the fields marshal and unmarshal to/from the admin API; the nil versions aren't of much interest.

kong/plugin_service_test.go
rainest: 9723ed58466bfc917b0f3909b39347812621a736

GGabriele (approved):

PR #190, opened 2022-07-13
feat(ci): license GH action

pmalek:
This PR aims to use https://github.com/Kong/kong-license#github-action to get the EE license for integration tests.

This will basically switch from a model that stores the license in Github Secrets (which needs to be periodically updated) to a synchronous pull on every job run.

Ideally this could be made as a shared step across jobs but for security (?) reasons Github doesn't send the job outputs from runners to Github Actions when it (magically?) detects that an output can contain a secret and in such a case we get:

```
Warning: Skip output 'license' since it may contain secret.
```


.github/workflows/integration-test-enterprise.yaml
```diff
@@ -40,6 +39,16 @@ jobs:
         # needed because the postgres container does not provide a healthcheck
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     steps:
+      # This will set KONG_LINCENSE_DATA environment variable.
+      # NOTE:
+      # This could be (?) made a shared step but for some reason it's impossible
+      # to override:
+      # Warning: Skip output 'license' since it may contain secret.
+      # on Github Actions when setting a job output.
+      - uses: Kong/kong-license@9fb64ad7be1ed2b121a70990783d7c0869e531d5
```
Tieske, line 48: I've wondered about pinning the version/commit. The action has a single functionality; setting the KONG_LICENSE_DATA env var. In this process there will be 2 moving parts;

1. the action code itself
2. the location where the license is stored

I would expect the latter to change over time, but that would be compensated by a change to the former (so transparent to the user). Now if we pin the commit of the action, then it would be counterproductive, since it would actually make it fail in that case.

@hutchic  wdyt?

hutchic: ideally pin but let dependabot PR updates. Especially for github actions that we don't own otherwise we risk getting supply chain attacked.

Tieske: This one we own. That said, we can pin this one, and see what happens if we ever change anything.

mflendrich (approved):

.github/workflows/integration-test-enterprise.yaml
pmalek: It was 17 days now :) 😅 I'm resolving this conversation and merging it. Let me know if there are any other concerns.

PR #197, opened 2022-08-11
chore: release 0.31

rainest:
Upgrades Go since we've been doing it for other projects.

Makes @achoimet's contribution from #192 available in a tag.


pmalek (approved):
Ideally we'd release in a separate PR but 🤷 

👍

shaneutt (approved):

jrsmroz (approved):

PR #205, opened 2022-08-22
fix - let the PluginId be optional like the create method.

achoimet:
Small fix for the PR : https://github.com/Kong/go-kong/pull/192

It seems that we let a small error within the PR, the PluginID was mandatory, sorry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [v0.31.1](#v0311)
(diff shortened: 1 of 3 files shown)


achoimet:
Hello @rainest, sorry to bother you, I fixed a small error from the previous MR.

rainest (changes requested):
Change itself looks fine. Could you update the test to include calls that use both the PUT and POST variant?

rainest (approved):

PR #207, opened 2022-09-01
feat: add support for Kong Gateway four digit versioning

mikefero:
This feature adds four digit versioning for Kong Gateway; OSS and enterprise. Starting with 3.0.0.0 Kong Gateway has dropped the "enterprise-edition" suffix from the version in both the CLI output and the version field from the root endpoint. Moving forward all four digit versions and any version ending with the "enterprise" or "enterprise-edition" suffix will be considered an enterprise edition.

This feature also allows for updated version range comparisons to use the fourth digit. This is beneficial when comparing against revision updates for Kong Gateway allowing for more granular specifications.

Note: This feature also removes pre-release and build metadata information from the underlying kong/semver implementation so that any pre-release can be compared to an actual version number.

### DIFF
--- a/kong/versioning.go
+++ b/kong/versioning.go
@@ -1,0 +1,157 @@
+package kong
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/kong/semver/v4"
+)
+
+// Version represents a three or four digit version.
+type Version struct {
+	// version represents a three of four digit version using the forked kong/semver
+	// library.
+	version semver.Version
+	// str represents the original version string when creating Version.
+	str string
+}
+
+// Range represents a range of versions which can be use to validate if a Version is valid
+// for a Range.
+type Range func(Version) bool
+
+// NewVersion creates a new instance of a Version.
+func NewVersion(versionStr string) (Version, error) {
+	version, err := semver.Parse(versionStr)
+	if err != nil {
+		return Version{}, fmt.Errorf("unable to create version: %w", err)
+	}
+
+	// Remove pre-release and build metadata versioning for finalized version comparisons
+	version.Pre = []semver.PRVersion{}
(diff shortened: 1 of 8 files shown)


mikefero:
**Note**: The enterprise tests are failing due to this PR coming from an external contribution. I don't have access on this repository to create PRs.

pmalek (approved):
🚀

kong/versioning.go
```diff
@@ -0,0 +1,157 @@
...
+// Version represents a three or four digit version.
+type Version struct {
+	// version represents a three of four digit version using the forked kong/semver
+	// library.
+	version semver.Version
+	// str represents the original version string when creating Version.
+	str string
+}
+
+// Range represents a range of versions which can be use to validate if a Version is valid
```
GGabriele, line 20:
```suggestion
// Range represents a range of versions which can be used to validate if a Version is valid
```

kong/versioning.go
```diff
@@ -0,0 +1,157 @@
...
+// for a Range.
+type Range func(Version) bool
+
+// NewVersion creates a new instance of a Version.
+func NewVersion(versionStr string) (Version, error) {
+	version, err := semver.Parse(versionStr)
+	if err != nil {
+		return Version{}, fmt.Errorf("unable to create version: %w", err)
```
GGabriele, line 28: does the error from `semver` include the initial string? Otherwise it may be useful to include it in the error message.

mikefero: It does, here is an example of the error message using `3.0` as the `versionStr`

```
unable to create version: Invalid version: '3.0'
```

The `Invalid version: '3.0'` comes from the underlying `kong/semver` library.

GGabriele (approved):
💯

PR #210, opened 2022-09-12
feat: add 'expression' and 'priority' fields to routes schema

GGabriele:


rainest:
Any idea how this is causing that 1.5 Enterprise failure? It doesn't look like there's any reason this should cause it, but it does appear consistent and unique to this change based on review of the other recent integration test runs.

rainest (approved):

PR #217, opened 2022-09-19
Add plugin to a route

achoimet:
Hello,
This PR to be able to add plugins at route level, same need as https://github.com/Kong/go-kong/pull/192 but at a lower granularity.


kong/plugin_service.go
```diff
@@ -210,6 +252,25 @@ func (s *PluginService) DeleteForService(ctx context.Context,
...
+	if isEmptyString(routeIDorName) {
+		return fmt.Errorf("routeIDorName cannot be nil")
+	}
+
+	endpoint := fmt.Sprintf("/routes/%v/plugins/%v", *routeIDorName, *pluginID)
+	_, err := s.sendRequest(ctx, nil, endpoint, "DELETE")
+	if err != nil {
+		return err
+	}
+	return err
```
pmalek, line 271:
```suggestion
	return nil
```

pmalek (commented):
Minor nit but overall this looks good

kong/plugin_service.go
```diff
@@ -210,6 +252,25 @@ func (s *PluginService) DeleteForService(ctx context.Context,
...
+	if isEmptyString(routeIDorName) {
+		return fmt.Errorf("routeIDorName cannot be nil")
+	}
+
+	endpoint := fmt.Sprintf("/routes/%v/plugins/%v", *routeIDorName, *pluginID)
+	_, err := s.sendRequest(ctx, nil, endpoint, "DELETE")
+	if err != nil {
+		return err
+	}
+	return nil
```
pmalek, line 271: This could have been simply `return err` but 🤷

pmalek (approved):

pmalek:
The integration tests were not designed to be run from forks, so we cannot really take their results into account.

Thanks for your contribution @achoimet !

PR #224, opened 2022-10-04
feat: add support to vaults

GGabriele:
https://docs.konghq.com/gateway/3.0.x/kong-enterprise/secrets-management/advanced-usage/


kong/vault_service_test.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestVaultsService(t *testing.T) {
+	RunWhenEnterprise(t, ">=3.0.0", RequiredFeatures{})
+	assert := assert.New(t)
+	require := require.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
```
pmalek, line 17: Similarly as in the previous PR: let's use `require` for those.

```suggestion
	require.NoError(err)
```

kong/vault_service_test.go
```diff
@@ -0,0 +1,207 @@
...
+	assert.NotNil(client)
```
pmalek, line 18: Ditto here (and elsewhere)

```suggestion
	require.NotNil(client)
```

kong/vault_service_test.go
```diff
@@ -0,0 +1,207 @@
...
+	assert.Equal(Configuration{"region": "us-east-2"}, createdVault.Config)
+
+	err = client.Vaults.Delete(defaultCtx, createdVault.ID)
+	assert.NoError(err)
+
+	_, err = client.Vaults.Create(defaultCtx, nil)
+	assert.NotNil(err)
+
+	_, err = client.Vaults.Update(defaultCtx, nil)
+	assert.NotNil(err)
```
pmalek, line 77: Do you mind adding error messages to those assertions? Something like `an update called with the same object as create shouldn't fail` (or similar that would fit here).

Without those I feel those are a bit without context.

Alternatively what you could do is to put those into subtests with proper names.

kong/vault_service_test.go
```diff
@@ -0,0 +1,207 @@
...
+			continue
+		}
+		expectedPrefixes = append(expectedPrefixes, *vault.Prefix)
+	}
+
+	for _, vault := range actual {
+		actualPrefixes = append(actualPrefixes, *vault.Prefix)
+	}
+
+	return (compareSlices(expectedPrefixes, actualPrefixes))
```
pmalek, line 206:
```suggestion
	return compareSlices(expectedPrefixes, actualPrefixes)
```

kong/vault_service.go
```diff
@@ -0,0 +1,158 @@
...
+// List fetches a list of Vaults in Kong.
+// opt can be used to control pagination.
+func (s *VaultService) List(ctx context.Context,
+	opt *ListOpt,
+) ([]*Vault, *ListOpt, error) {
+	data, next, err := s.client.list(ctx, "/vaults", opt)
+	if err != nil {
+		return nil, nil, err
+	}
+	var vaults []*Vault
```
pmalek, line 125:
```suggestion
	vaults := make([]*Vault, 0, len(data))
```

kong/vault_service.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, object := range data {
+		b, err := object.MarshalJSON()
```
pmalek, line 127: It seems that the underlying type of [`json.RawMessage`](https://pkg.go.dev/encoding/json#RawMessage) is `[]byte` which means that it should be possible to "just use it" as `json.Unmarshal()` parameter. This would save us the unnecessary marshalling.

Can't test it now because it seems I need to do some shenanigans to start kong locally (manually) to do so.

pmalek (changes requested):

kong/vault_service.go
GGabriele: Good catch!

kong/vault_service_test.go
GGabriele: Sure, I agree these asserts checking the calls error out are a bit obscure without hints.

kong/vault_service_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	require.True(compareVaults(t, vaults, vaultsFromKong))
+
+	vaults, err = client.Vaults.ListAll(defaultCtx)
+	require.NoError(err)
+	require.NotNil(vaults)
+	require.Equal(3, len(vaults))
+
+	for i := 0; i < len(vaults); i++ {
+		require.NoError(client.Vaults.Delete(defaultCtx, vaults[i].ID))
```
pmalek, line 186: Oh, wait. Is that meant to clean up the environment? If that's the case I suggest to use something like

```go
t.Cleanup(func(){
  // Note the assert here as we might want more logic to be run as part of the cleanup,
  // regardless of the return value of removing this particular object.
  assert.NoError(client.Vaults.Delete(defaultCtx, vaults[i].ID))
})
```

called just after a successful `.Create()`.

kong/vault_service_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestVaultsService(t *testing.T) {
```
pmalek, line 11: Can we use the cleanup pattern here as well? If any assertion fails (or the test panics between the vault creation and the deletion itself) we'll be left with leftovers in kong's config wouldn't we?

PR #231, opened 2022-10-24
feat: support JSON schemas in the FillEntityDefaults function

GGabriele:
FillEntityDefaults can be used to inject Kong entity defaults into a Kong entity struct. This function achieves this by getting entity schemas from the Admin API, flattening the returned schemas and merging these with the input schemas.

Kong CP only works with Lua schemas, while Konnect also supports JSON schemas.

This commit adds support to filling entity defaults from JSON schemas.

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -248,7 +248,9 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 // and flattens it, turning it into a map that can be more easily unmarshalled
 // into proper entity objects.
 //
-// Sample input:
+// This supports both Lua and JSON schemas.
+//
+// Sample Lua schema input:
 //
 //	{
 //		"fields": [
@@ -274,6 +276,32 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 //	  ...
 //	}
 //
+// Sample JSON schema input:
+//
+//	{
+//		"properties": [
+//	        {
+//	            "algorithm": {
+//	                "default": "round-robin",
+//	                "enum": ["consistent-hashing", "least-connections", "round-robin"],
+//	                "type": "string"
+//	            }
+//	        }, {
+//	            "hash_on": {
+//	                "default": "none",
+//	                "enum": ["none", "consumer", "ip", "header", "cookie"],
+//	                "type": "string"
+//	            }
+//	        }, {
+//	            "hash_fallback": {
+//	                "default": "none",
+//	                "enum": ["none", "consumer", "ip", "header", "cookie"],
+//	                "type": "string"
+//	            }
+//	        },
+//	  ...
+//	}
+//
 // Sample output:
 //
 //	{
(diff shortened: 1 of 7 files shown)


kong/utils_test.go
```diff
@@ -567,6 +570,363 @@ func TestFillUpstreamsDefaults(T *testing.T) {
 	}
 }
 
+func getJSONSchemaFromFile(filename string) Schema {
+	jsonFile, err := os.Open(filename)
+	if err != nil {
+		panic(err)
```
pmalek, line 576: Ouch, why 😢 ? Why not pass a `*testing.T` and `require.NoError(t, err)`?

kong/utils_test.go
```diff
@@ -567,6 +570,363 @@ func TestFillUpstreamsDefaults(T *testing.T) {
...
+	}
+	defer jsonFile.Close()
+
+	var schema Schema
+	byteValue, err := io.ReadAll(jsonFile)
+	if err != nil {
+		panic(err)
```
pmalek, line 583: ditto

kong/utils_test.go
```diff
@@ -567,6 +570,363 @@ func TestFillUpstreamsDefaults(T *testing.T) {
...
+	}
+	if err := json.Unmarshal(byteValue, &schema); err != nil {
+		panic(err)
```
pmalek, line 586: ditto

kong/utils_test.go
```diff
@@ -567,6 +570,363 @@ func TestFillUpstreamsDefaults(T *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			u := tc.upstream
+			if err := FillEntityDefaults(u, schema); err != nil {
+				t.Errorf(err.Error())
+			}
```
pmalek, line 717:
```suggestion
			require.NoError(t, FillEntityDefaults(u, schema))
```

Simpler and more succint?

pmalek (changes requested):

kong/utils_test.go
GGabriele: I like simpler :D

kong/utils_test.go
```diff
@@ -567,6 +571,355 @@ func TestFillUpstreamsDefaults(T *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			target := tc.target
+			require.NoError(t, FillEntityDefaults(target, schema))
+			if diff := cmp.Diff(target, tc.expected); diff != "" {
+				t.Errorf(diff)
+			}
```
pmalek, line 918: How about:

```suggestion
			require.Equal(t, tc.expected, target)
```

?

kong/utils_test.go
```diff
@@ -567,6 +571,355 @@ func TestFillUpstreamsDefaults(T *testing.T) {
...
+
+	var schema Schema
+	byteValue, err := io.ReadAll(jsonFile)
+	if err != nil {
+		require.NoError(t, err)
+	}
+	if err := json.Unmarshal(byteValue, &schema); err != nil {
+		require.NoError(t, err)
+	}
+	return schema
```
pmalek, line 589: To consider:

```suggestion
	jsonFile, err := os.Open(filename)
	require.NoError(t, err)
	defer jsonFile.Close()

	var schema Schema
	require.NoError(t, json.NewDecoder(jsonFile).Decode(&schema))
	return schema
```

An advantage of the above is that you don't need to read the whole file into memory at once.

kong/utils.go
```diff
@@ -283,7 +311,49 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+
+		ftype := value.Get("type")
+		// when type==object and additionalProperties==false, the object
+		// represents either a foreign relationship or a map entry.
+		// In both cases, defaults don't need to be injected.
+		additionalProperties := value.Get("additionalProperties")
+		if ftype.String() == "object" &&
+			(!additionalProperties.Exists() ||
+				(additionalProperties.Exists() &&
+					additionalProperties.Bool())) {
```
pmalek, line 339: Shouldn't this be `!additionalProperties.Bool()` ? 🤔

GGabriele: Nope, we don't want to execute the following logic when `additionalProperties` is `false`

kong/utils_test.go
GGabriele: Nice, thanks. Fixed!

kong/utils_test.go
GGabriele: Yep!

pmalek (approved):

PR #237, opened 2022-12-02
feat: include raw response bodies in APIErrors

rainest:
This adds raw error response bodies to APIErrors. This is a breaking change, as it modifies the public `kong.NewAPIError()` function signature, adding a new `raw []byte` parameter for the raw body.

go-kong does not provide universal coverage of all Kong endpoints, and does not have rich error handling for all endpoints it does cover. Currently, go-kong reads the body of any admin API error response (consuming it, so nothing downstream can read it) and extracts the `message` field for inclusion in the APIError it returns. It is not possible to perform further parsing on error bodies if desired.

Adding the raw response allows downstream clients to handle their own error parsing, primarily for ad-hoc requests where downstream builds its own requests and runs it via a go-kong client. I think this is also necessary for further parsing inside go-kong (I don't think raw bodies were available to go-kong services either), but need to confirm.

---

This came up in the context of KIC handling `POST /config` requests. go-kong does not provide a service for `/config` and does not provide error parsing for it as such. `message` isn't useful for this endpoint because it's a compressed version of the more machine-readable `fields` field those error responses.

### DIFF
--- a/kong/error.go
+++ b/kong/error.go
@@ -10,12 +10,21 @@ import (
 type APIError struct {
 	httpCode int
 	message  string
+	raw      []byte
 }
 
 func NewAPIError(code int, msg string) *APIError {
 	return &APIError{
 		httpCode: code,
 		message:  msg,
+	}
+}
+
+func NewAPIErrorWithRaw(code int, msg string, raw []byte) *APIError {
+	return &APIError{
+		httpCode: code,
+		message:  msg,
+		raw:      raw,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


rainest:
Undrafting this in prep for expected gateway changes in FTI-4366. Although the expected format still isn't set in stone, getting this into go-kong will let us iterate rapidly on future gateway changes, since we'll always have the raw response available.

kong/error.go
```diff
@@ -10,12 +10,14 @@ import (
 type APIError struct {
 	httpCode int
 	message  string
+	raw      []byte
 }
 
-func NewAPIError(code int, msg string) *APIError {
+func NewAPIError(code int, msg string, raw []byte) *APIError {
```
pmalek, line 16: How about we don't change this but include a second "constructor" called `NewAPIErrorWithBody` or `NewAPIErrorWithRaw(Body)` and use that in `NewAPIError`?

This way we'll avoid passing the third param `[]byte{}` where it's not needed.

czeslavo: 👍 + we do not break the API.

czeslavo (approved):

PR #245, opened 2022-12-08
ci: don't run ee tests when license unavailable

pmalek:
Since we can't reliably and securely give outside contributors access to our secrets we can't run the EE tests for those PRs.

This PR will allow us to accept patches contributed from outside of our org without enterprise tests failing.

### DIFF
--- a/.github/workflows/integration-test-enterprise.yaml
+++ b/.github/workflows/integration-test-enterprise.yaml
@@ -28,6 +28,18 @@ jobs:
       KONG_VERSION: ${{ matrix.kong_version }}
       KONG_ANONYMOUS_REPORTS: "off"
       KONG_ADMIN_TOKEN: kong
+      # We don't want to run enterprise tests for contributors whose workflow runs
+      # do not have access to enterprise license (here, through the means of
+      # kong-license action) so don't run this job when the required secret is not
+      # available.
+      # Instead of making it conditional on job level we need to job.env because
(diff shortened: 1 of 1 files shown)


pmalek:
> Is there a way you can skip this [at the job level](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution)? I expect you can't check the env value, because the job is responsible for setting it, but you may be able to evaluate `secrets.PULP_PASSWORD == ''` directly, or move the license step into separate job and then test/use [its output](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs)

As per my comment in the code : I've tried to do something similar but failed because the env and secrets context are not available at that level. We can do what you suggest sure. Feel free to submit a proposal :)

rainest:
https://github.com/Kong/go-kong/pull/247 takes the separate job route.

However, did we plan to make the "all enterprise" job a required check? I don't think there's a way to make that simultaneously succeed if all test jobs succeed or if all test jobs are skipped.

pmalek:
> https://github.com/Kong/go-kong/pull/247 takes the separate job route.
> 
> However, did we plan to make the "all enterprise" job a required check? I don't think there's a way to make that simultaneously succeed if all test jobs succeed or if all test jobs are skipped.

I'm afk now but this works AFAIR (i.e if you skip all steps then the result is success)

pmalek:
Closed in favor of #247

PR #247, opened 2022-12-08
chore: add pre-flight license check

rainest:
[Pre-flight](https://github.com/Kong/go-kong/pull/245#pullrequestreview-1210533100) version of #245.

This PR actually runs the tests, since it's local.

https://github.com/Kong/go-kong/pull/246 is from a fork, and does not.

This does explicitly skip `test-enterprise-passed`. We don't have this currently required, but if we do I'm not sure if there's a way to make it pass like the skipped steps.

### DIFF
--- a/.github/workflows/integration-test-enterprise.yaml
+++ b/.github/workflows/integration-test-enterprise.yaml
@@ -10,8 +10,25 @@ on:
     - '*'
 
 jobs:
+  secret-available:
+    outputs:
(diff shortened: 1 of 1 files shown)


pmalek (approved):

pmalek:
I like this one better since we can put the logic in one place and the CI will still allow us to merge even though the `test-enterprise-passed` check is required 👍 

Screenshot 2022-12-09 at 09 55 15

PR #252, opened 2022-12-14
feat: add Client.ReloadDeclarativeConfig()

pmalek:
This PR adds possibility to send declarative configs to Kong [via `/config`](https://docs.konghq.com/gateway/latest/admin-api/#reload-declarative-configuration).

It also makes `Client.NewRequestRaw()` more "flexible" ( I guess? ) by checking the type of the accepted argument and calls `json.Marshal()` only when the provided argument is not `string`, `[]byte` or `io.Reader`.

---

Unfortunately we cannot accept proper types as parameters because [deck already imports go-kong](https://github.com/Kong/deck/blob/be0eaa0835c8ca90fb5944481defe6ac534a41be/go.mod#L22) so using e.g. [`KongRawState`](https://github.com/Kong/deck/blob/0df3f51cb93669eaf661051f292eb176923c0e88/utils/types.go#L27-L56) or [`file.Content`](https://github.com/Kong/deck/blob/0df3f51cb93669eaf661051f292eb176923c0e88/file/types.go#L660-L682) is not possible because that would cause a circular dependency.

### DIFF
--- a/kong/client.go
+++ b/kong/client.go
@@ -39,6 +39,7 @@ type Client struct {
 	workspace               string       // Do not access directly. Use Workspace()/SetWorkspace().
 	workspaceLock           sync.RWMutex // Synchronizes access to workspace.
 	common                  service
+	Configs                 AbstractConfigService
 	Consumers               AbstractConsumerService
 	Developers              AbstractDeveloperService
 	DeveloperRoles          AbstractDeveloperRoleService
@@ -127,6 +128,7 @@ func NewClient(baseURL *string, client *http.Client) (*Client, error) {
 	kong.defaultRootURL = url.String()
 
 	kong.common.client = kong
+	kong.Configs = (*ConfigService)(&kong.common)
 	kong.Consumers = (*ConsumerService)(&kong.common)
 	kong.Developers = (*DeveloperService)(&kong.common)
 	kong.DeveloperRoles = (*DeveloperRoleService)(&kong.common)
--- a/kong/config_service.go
+++ b/kong/config_service.go
@@ -1,0 +1,65 @@
+//nolint:lll
+package kong
+
+import (
+	"context"
+	"fmt"
+	"io"
+)
+
+// AbstractConfigService handles Config in Kong.
+type AbstractConfigService interface {
+	// ReloadDeclarativeRawConfig sends out the specified config to configured Admin
+	// API endpoint using the provided reader which should contain the JSON
+	// serialized body that adheres to the configuration format specified at:
+	// https://docs.konghq.com/gateway/latest/production/deployment-topologies/db-less-and-declarative-config/#declarative-configuration-format
+	ReloadDeclarativeRawConfig(ctx context.Context, config io.Reader, checkHash bool) error
+}
+
+// ConfigService handles Config in Kong.
+type ConfigService service
+
+// ReloadDeclarativeRawConfig sends out the specified config to configured Admin
+// API endpoint using the provided reader which should contain the JSON
+// serialized body that adheres to the configuration format specified at:
+// https://docs.konghq.com/gateway/latest/production/deployment-topologies/db-less-and-declarative-config/#declarative-configuration-format
+func (c *ConfigService) ReloadDeclarativeRawConfig(
+	ctx context.Context,
+	config io.Reader,
+	checkHash bool,
+) error {
+	type sendConfigParams struct {
+		CheckHash int `url:"check_hash"`
+	}
+	var checkHashI int
+	if checkHash {
+		checkHashI = 1
+	}
+	req, err := c.client.NewRequest("POST", "/config", sendConfigParams{CheckHash: checkHashI}, config)
+	if err != nil {
+		return fmt.Errorf("creating new HTTP request for /config: %w", err)
+	}
+
+	resp, err := c.client.DoRAW(ctx, req)
+	if err != nil {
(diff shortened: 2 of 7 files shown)


kong/client.go
```diff
@@ -363,3 +363,37 @@ func (c *Client) RootJSON(ctx context.Context) ([]byte, error) {
 
 	return body, nil
 }
+
+// SendConfig sends out the specified config to configured Admin API endpoint
+func (c *Client) SendConfig(ctx context.Context, config Configuration) error {
```
czeslavo, line 368: Let's consider naming it `ReloadDeclarativeConfig` as this is how it's named in the official docs of the Admin API: https://docs.konghq.com/gateway/latest/admin-api/#reload-declarative-configuration. 

Probably a comment telling that it's valid only for the DB-less mode would be useful here.

kong/client.go
```diff
@@ -363,3 +363,37 @@ func (c *Client) RootJSON(ctx context.Context) ([]byte, error) {
...
+	type sendConfigParams struct {
+		CheckHash int `url:"check_hash"`
+	}
+	req, err := c.NewRequest("POST", "/config", sendConfigParams{CheckHash: 1}, config)
```
czeslavo, line 372: In the case of a public library, I think that would make sense to allow passing the `CheckHash` param to the method.

kong/client.go
pmalek: Done

kong/config_service.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+	"fmt"
+	"io"
+)
+
+// AbstractConfigService handles Config in Kong.
+type AbstractConfigService interface {
+	ReloadDeclarativeConfig(ctx context.Context, config any, checkHash int) error
```
czeslavo, line 11: I'd suggest making `checkHas` a boolean. Integer allows a user to pass an arbitrary number while it doesn't really make sense in such a high-level API I think - it can be hidden as a detail of the HTTP API call.

czeslavo, line 11: Would be great if we could somehow narrow down the type of `config`, ideally making it a strongly typed configuration 🤔 Allowing to pass any type does make this method very easy to break. 

If that's not possible to make it a strong type, let's maybe at least leave a comment describing what we're actually expecting it to be (e.g. "declarative config request body").

czeslavo, line 11: I think it would make sense to have the comment documenting the method on the interface level as well.

pmalek: I thought about that but there are a couple of points that make this "challenging":

- you can't add type parameters to methods to something like this:

  ```
  type configConstraint interface {
	  []byte | string | Configuration | io.Reader
  }
  ```

  is not allowed. You either specify a union of structs (any of those) and add an interface **which also should be fulfilled** or use 1 of those constraints. Can't make it an alternative. https://github.com/golang/go/issues/49054#issuecomment-946337356

- [`NewRequestRaw()`](https://github.com/Kong/go-kong/blob/d5120140735c43bb5f14f6cd2cb0cbe9f8d76d77/kong/request.go#L14) accepts `interface{}` and is widely used (and exported) so we can't just will nilly change this (or can we? we're pre v1 :) ). This is not a problem per se but would be a nice improvement to have the whole call stack guarded by type constraints.

---

PTAL at 535869c, I've added a second method which allows sending the config as `Configuration` whereas the original was adopted to accept `io.Reader`. I believe that's the best of 2 worlds for now.

czeslavo: Couldn't we leave just the strongly typed method? Is a reason for leaving the loosely typed `ReloadDeclarativeConfig` that we (in KIC) don't use the strong type when sending the config?

pmalek: It's a bit more nuanced 🤯 

In KIC we use [`*file.Content`](https://github.com/Kong/kubernetes-ingress-controller/blob/aac834bd06a89d488f61d927fb5541413e65499b/internal/dataplane/sendconfig/sendconfig.go#L125) from [deck](https://github.com/Kong/deck/blob/be0eaa0835c8ca90fb5944481defe6ac534a41be/file/types.go#L660). We can't use it in `go-kong` because that would create a circular dependency ([deck already imports go-kong](https://github.com/Kong/deck/blob/be0eaa0835c8ca90fb5944481defe6ac534a41be/go.mod#L22)).

Ideally this would be the type we use there but 🤷 I have no idea (other than using something else like `map[string]interface{}` :( or copy pasting) how to approach this properly.

Hence why I left the `io.Reader` as an option to pass serialized version of `*file.Content`, `Configuration`, `map[string]interface{}` or anything else if the user's  brave enough.

czeslavo: Hmm, got it. So maybe just to make the distinction between these two more favoring the strongly typed one, we could rename them as follows:
- `ReloadDeclarativeConfig` -> `ReloadRawDeclarativeConfig` (the Raw part should be discouraging usage for new users)
- `ReloadDeclarativeConfigWithSchema` -> `ReloadDeclarativeConfig` (the right ™️ one)
WDYT?

czeslavo: As we've discovered `kong.Configuration` is not really a type for the whole kong configuration, but a single plugin, I propose to keep only the `ReloadRawDeclarativeConfig` accepting an `io.Reader` for now.

pmalek: Done.

czeslavo (approved):

PR #255, opened 2022-12-16
feat: add RootBaseURL() to Client

pmalek:
This PR adds `.RootBaseURL()` to `Client` which allows the user to get the configured URL.


kong/client_test.go
```diff
@@ -199,3 +199,23 @@ func TestTestWorkspace(T *testing.T) {
...
+func TestBaseRootURL(t *testing.T) {
+	t.Run("default", func(t *testing.T) {
+		client, err := NewClient(nil, nil)
+		require.NoError(t, err)
+		require.NotNil(t, client)
+
+		require.Equal(t, client.BaseRootURL(), "http://localhost:8001")
+	})
+
+	// Don't test setting the URL via env because that can mess up other tests.
```
czeslavo, line 212: We can consider using `t.SetEnv` (it ensures that the test using it is not parallel so should be safe to use): 
https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/testing/testing.go;l=1287

CHANGELOG.md
```diff
@@ -50,6 +50,8 @@
   [#231](https://github.com/Kong/go-kong/pull/231)
 - Add possibility to client to send declarative configs via `ReloadDeclarativeRawConfig()`
   [#252](https://github.com/Kong/go-kong/pull/252)
+- Add `RootBaseURL()` to Client #255
```
czeslavo, line 53:
```suggestion
- Add `RootBaseURL()` to Client
```

kong/client_test.go
pmalek: Added. Thanks, one learns something every day!

CHANGELOG.md
```diff
@@ -50,6 +50,12 @@
   [#231](https://github.com/Kong/go-kong/pull/231)
 - Add possibility to client to send declarative configs via `ReloadDeclarativeRawConfig()`
   [#252](https://github.com/Kong/go-kong/pull/252)
+<<<<<<< HEAD
```
mlavacca, line 53: some rebase leftover :)

CHANGELOG.md
```diff
@@ -50,6 +50,12 @@
...
+- Add `RootBaseURL()` to Client
+=======
+- Add `RootBaseURL()` to Client #255
```
mlavacca, line 56: looks like there is a misnaming here, as the method is called `BaseRootURL`, not `RootBaseURL`

kong/client.go
```diff
@@ -365,3 +365,7 @@ func (c *Client) RootJSON(ctx context.Context) ([]byte, error) {
 
 	return body, nil
 }
+
+func (c *Client) BaseRootURL() string {
```
mlavacca, line 369: WDYT about naming it `DefaultRootURL` to make it clear this is a getter for `c. defaultRootURL`?

CHANGELOG.md
pmalek: Thanks for staying vigilant!

kong/client.go
pmalek: I somehow don't like the `default` naming in here 🤔 Reading this as a user I'd get a feeling that I'm getting a default value that could be used if nothing else was specified.

czeslavo: I'd say the `defaultRootURL` field naming is misleading. It should be named IMO `rootURL` instead as it's not always falling back to the default one.

pmalek: Vey much so. Let it be that we fix this as we go. PTAL

czeslavo: BTW the part calculating it in the constructor could be extracted to a separate function/method to make it more obvious. 🤔

CHANGELOG.md
```diff
@@ -50,6 +50,8 @@
   [#231](https://github.com/Kong/go-kong/pull/231)
 - Add possibility to client to send declarative configs via `ReloadDeclarativeRawConfig()`
   [#252](https://github.com/Kong/go-kong/pull/252)
+- Add `RootBaseURL()` to Client
```
mlavacca, line 53:
```suggestion
- Add `BaseRootURL()` to Client
```

pmalek: Already pushed this, thanks!

mlavacca (approved):

PR #258, opened 2022-12-22
fix: correctly fill plugins with arbitrary map fields

GGabriele:
Kong plugin schemas support different type of fields, including maps. Moreover, these can either be arbitrary maps, or maps with their own subschema.

go-kong already supports well the second case, but it falls short with arbitrary maps. An example of arbitrary map field is the `custom_fields_by_lua` from the `http-log` plugin, which is defined as follows:

```
{
  "custom_fields_by_lua": {
    "keys": {
      "len_min": 1,
      "type": "string"
    },
    "type": "map",
    "values": {
      "len_min": 1,
      "type": "string"
    }
  }
}
```

This commit makes sure that go-kong is able to correctly handle such schemas as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [v0.34.1](#v0341)
 - [v0.34.0](#v0340)
 - [v0.33.0](#v0330)
 - [v0.32.0](#v0320)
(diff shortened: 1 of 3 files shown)


aboudreault (approved):

GGabriele:
@Kong/team-k8s this PR is approved (but merge is blocked) and decK is already using this code. Can someone take a look please?

pmalek (approved):

pmalek:
> @Kong/team-k8s this PR is approved (but merge is blocked) and decK is already using this code. Can someone take a look please?

Done :)

PR #263, opened 2023-01-10
test: remove TestRunWhenEnterprise

### LINKED ISSUE #212: TestRunWhenEnterprise broken after 3.x
## Problem statement

3.x removes the `-enterprise` version suffix, which this test was looking for. There's not an immediately obvious way to rework the test such that it isn't simply confirming that the new IsKongGatewayEnterprise returns the same value for the same input.

## Possible solution

Make https://github.com/Kong/go-kong/blob/ae8c79a632a47a6dc957ca340b66ebce0a2b8461/kong/versioning.go#L106-L109 take into account that if the GW version contains 4 segments then it's enterprise.

E.g. `3.0.2.0` is an enterprise version. `3.0.2` is not.

## Acceptance criteria

- [ ] [`TestRunWhenEnterprise`](https://github.com/Kong/go-kong/blob/9ce90cdb1a5c25d0b7e5eb6e5f1280ee799b1cb4/kong/client_test.go#L118) runs on GW versions 3.0 and up

pmalek:
Remove `TestRunWhenEnterprise` since all it's doing is checking wether the version information retrieved via `client.Root()` is an enterprise version when `RunWhenEnterprise()` check passes which uses the same methods underneath to check whether the used version of Kong Gateway is indeed enterprise 🤯 

Fixes #212

### DIFF
--- a/kong/client_test.go
+++ b/kong/client_test.go
@@ -115,23 +115,6 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func TestRunWhenEnterprise(T *testing.T) {
-	// TODO refactor this to test that a version is Enterprise without relying on the IsKongGatewayEnterprise function
-	// that this calls https://github.com/Kong/go-kong/issues/212
-	RunWhenEnterprise(T, ">=0.33.0 <3.0.0", RequiredFeatures{})
-	assert := assert.New(T)
-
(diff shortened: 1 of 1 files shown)


mlavacca (approved):
I feel dirty in approving a unit test deletion, but if it's useless 🤷

PR #264, opened 2023-01-12
feat: add IsForbiddenErr error checking for 403s

GGabriele:


kong/error_test.go
```diff
@@ -49,3 +49,18 @@ func TestAPIError_Code(T *testing.T) {
 	require.True(ok)
 	assert.True(kongErr.Code() == 404)
 }
+
+func TestIsForbiddenErrE2E(T *testing.T) {
+	T.Skip()
```
GGabriele, line 54: This test is expected to work if no license is present or if that's expired.

@pmalek do you have any tips to make this work in the go-kong CI (which always has a valid license)? Or, even better, any other option to trigger a 403 coming to mind?

pmalek: It might be a bit more involved but using a custom roundTripper like https://github.com/Kong/go-kong/blob/d8d2d9de7467a1870080a60b23cf059e24868bcf/kong/utils.go#L81-L103 which you can mock response of would fit here I guess.

`RunWhenEnterprise` unfortunately doesn't allow you to provide a custom client (where you'd use that roundTripper) but you could expand its parameters list 🤷.

GGabriele: good point. I modified this with a custom client, thanks for the tip!

kong/error_test.go
```diff
@@ -49,3 +51,36 @@ func TestAPIError_Code(T *testing.T) {
...
+	httpClient := &http.Client{
+		Transport: forbiddenTransport,
+	}
+	// if KONG_ADMIN_TOKEN is set, a different custom transport would be used
+	T.Setenv("KONG_ADMIN_TOKEN", "")
+	adminToken := os.Getenv("KONG_ADMIN_TOKEN")
+	os.Unsetenv("KONG_ADMIN_TOKEN")
+	defer func() {
+		os.Setenv("KONG_ADMIN_TOKEN", adminToken)
+	}()
```
pmalek, line 70: `t.Setenv()` will automatically restore the env to previous value using `t.Cleanup()`

```suggestion
	// if KONG_ADMIN_TOKEN is set, a different custom transport would be used
	T.Setenv("KONG_ADMIN_TOKEN", "")
```

kong/error_test.go
```diff
@@ -49,3 +51,36 @@ func TestAPIError_Code(T *testing.T) {
...
+
+	client, err := NewTestClient(nil, httpClient)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 74: Can we use `require` instead of `assert` here? This will stop the test execution when these statements fail.

```suggestion
	require.NoError(T, err)
	require.NotNil(T, client)
```

kong/error_test.go
```diff
@@ -49,3 +51,36 @@ func TestAPIError_Code(T *testing.T) {
...
+
+	_, err = client.Consumers.ListAll(defaultCtx)
+	assert.NotNil(err)
```
pmalek, line 77:
```suggestion
	require.Error(T, err)
```

This will pretty print the error and has a more specific use case than `NotNil()`

kong/error_test.go
GGabriele: I tried that and it didn't work, but let's give it another try

GGabriele: Yeah it doesn't work: https://github.com/Kong/go-kong/actions/runs/3910665439/jobs/6683126778

The `os.LookupEnv("KONG_ADMIN_TOKEN")` returns `true` even if the env value is set to an empty string. Unfortunately, golang testing lib doesn't have the equivalent of `Unsetenv` :\

GGabriele: Let me know if something like this would be okay for you https://github.com/Kong/go-kong/pull/264/commits/102506c5fd1cfcad2ac9bc84ff2b78b991b24718

pmalek: If that works then sure 👍

pmalek (approved):

PR #265, opened 2023-01-19
feat: add consumer group tags

aboudreault:


kong/consumer_group_test.go
```diff
@@ -55,6 +55,35 @@ func TestConsumerGroupsService(t *testing.T) {
...
+	require.Equal(cg.Tags, createdConsumerGroup.Tags)
+
+	createdConsumerGroup.Tags = StringSlice("tag1", "tag2", "tag3")
+	updatedConsumerGroup, err := client.ConsumerGroups.Update(defaultCtx, createdConsumerGroup)
+	require.NoError(err)
+	assert.NotNil(updatedConsumerGroup)
+	assert.Equal(updatedConsumerGroup.Tags, createdConsumerGroup.Tags)
+
+	err = client.Vaults.Delete(defaultCtx, updatedConsumerGroup.ID)
+	require.NoError(err)
```
pmalek, line 84: Do you mind moving this just below the 

```
	createdConsumerGroup, err := client.ConsumerGroups.Create(defaultCtx, cg)
	require.NoError(err)
```

and putting this into a `t.Cleanup()`? This way regardless of what happens (e.g. a `require` assertion breaks) we get that cleaned up.

aboudreault: sure, fixed.

kong/consumer_group_test.go
```diff
@@ -55,6 +55,36 @@ func TestConsumerGroupsService(t *testing.T) {
...
+	cg := &ConsumerGroup{
+		Name: String("foo"),
+		Tags: StringSlice("tag1", "tag2"),
+	}
+
+	createdConsumerGroup, err := client.ConsumerGroups.Create(defaultCtx, cg)
+	require.NoError(err)
+	t.Cleanup(func() {
+		err = client.ConsumerGroups.Delete(defaultCtx, createdConsumerGroup.ID)
+		require.NoError(err)
```
pmalek, line 76:
```suggestion
		assert.NoError(err)
```

I realize that there's no other deferred functions or cleanups but better to use `assert` here just in case there's something added to run on test teardown to not prevent the further cleanup being done.

pmalek (approved):

pmalek:
There's one more thing I almost forgot: do you mind adding a changelog entry for this? 🙏

pmalek (approved):

PR #267, opened 2023-01-19
fix: unbreak NewAPIErrorWithRaw and NewAPIError-related tests

rainest:
In retrospect from https://github.com/Kong/go-kong/pull/237, we actually do want to break the exported `NewAPIError` function. AFAICT we use this in [one place only](https://github.com/Kong/kubernetes-ingress-controller/blob/e3c862dee98e9e6f2bb4be582119a938f5edebf3/internal/dataplane/sendconfig/sendconfig_test.go#L30), and that place is a test within KIC that we can change trivially (searched through deck and don't see any use there).

Introducing the separate one apparently [broke all sorts of other things](https://github.com/Kong/go-kong/actions/runs/3962344750) in go-kong's _internal_ test expectations and apparently this repo is not configured to block auto-merge on that :sob: 

https://github.com/Kong/go-kong/actions/runs/3962834113/jobs/6790000801 suggests that the skipped ` test-passed` actually satisfies the branch protection requirements, somehow.


pmalek:
I think I'd just try to address the failing test rather than doing that.

Or perhaps change the API to accept options like `WithBody([]byte)`? Passing `nil` in places where an argument is "not really necessary" feels like a design flaw to me.

rainest:
Yah, sorry, my bad, was overthinking it and mixed up the tests.

kong/response.go
```diff
@@ -37,3 +37,12 @@ func hasError(res *http.Response) error {
 	body, _ := io.ReadAll(res.Body) // TODO error in error?
 	return NewAPIError(res.StatusCode, messageFromBody(body))
 }
+
+func hasErrorRaw(res *http.Response) error {
+	if res.StatusCode >= 200 && res.StatusCode <= 399 {
+		return nil
+	}
+
+	body, _ := io.ReadAll(res.Body) // TODO error in error?
```
pmalek, line 46: Can we just wrap the error from this and put into the return value propagating whatever was returned by `io.ReadAll`? Otherwise this will get lost. Rarely but still, it can happen.

kong/response.go
```diff
@@ -34,6 +34,21 @@ func hasError(res *http.Response) error {
...
+	}
 	return NewAPIError(res.StatusCode, messageFromBody(body))
 }
+
+func hasErrorRaw(res *http.Response) error {
+	if res.StatusCode >= 200 && res.StatusCode <= 399 {
+		return nil
+	}
+
+	body, err := io.ReadAll(res.Body) // TODO error in error?
```
pmalek, line 49: Do we need this todo? If so let's create an issue for that, tracking it. Otherwise let's remove it

pmalek (approved):

kong/response.go
rainest: nope

PR #271, opened 2023-01-30
feat: return error response bodies in config updates and fold config service into Kong client

rainest:
Returns Kong `POST /config` bodies from `ReloadDeclarativeRawConfig()`.

Folds `ConfigService` into `kong.Client`. go-kong services generally interact with a single distinct Kong entity (a route, service, consumer, etc.) whereas `POST /config` sends a blob of various kinds of entities in a single request.

Releases 0.37.0. Note that these are both breaking changes, but, the config service was only recently added and only used in one place in KIC, and we're updating that part of KIC (in part to make use of the now-returned response body) anyway, so shouldn't be cause for concern.


kong/config_service_test.go
```diff
@@ -76,7 +76,7 @@ func TestConfigService(t *testing.T) {
 			b, err := json.Marshal(tt.config)
 			require.NoError(t, err)
 
-			if err := client.Configs.ReloadDeclarativeRawConfig(ctx, bytes.NewBuffer(b), true); (err != nil) != tt.wantErr {
+			if err, _ := client.Configs.ReloadDeclarativeRawConfig(ctx, bytes.NewBuffer(b), true); (err != nil) != tt.wantErr {
```
pmalek, line 79:
```suggestion
			if _, err := client.Configs.ReloadDeclarativeRawConfig(ctx, bytes.NewBuffer(b), true); (err != nil) != tt.wantErr {
```

whoops

pmalek (commented):
Since we're running tests in this repo against actual Kong instances we should add a tests for this functionality (guarded by a version check)

rainest:
> Since we're running tests in this repo against actual Kong instances we should add a tests for this functionality (guarded by a version check)

We are stuck for lack of an actual Kong version that includes this, however. https://github.com/Kong/kong/pull/10161 isn't in `master` yet.

pmalek:
> > Since we're running tests in this repo against actual Kong instances we should add a tests for this functionality (guarded by a version check)
> 
> We are stuck for lack of an actual Kong version that includes this, however. [Kong/kong#10161](https://github.com/Kong/kong/pull/10161) isn't in `master` yet.

Yup, hence my proposing to guard this by a version check (when we know which version it will be).

rainest:
On further thought, is there anything we want to test re the body? go-kong isn't (yet) responsible for anything beyond receiving and returning it. All more complex logic is currently downstream (i.e. the error parsing in https://github.com/Kong/kubernetes-ingress-controller/pull/3446).

go-kong could be responsible for error parsing and probably should be responsible for it long-term, but at present we only have the existing Kong error format, which we effectively cannot parse, and the as-yet unreleased https://github.com/Kong/kong/pull/10161 format, which we can parse and do parse in KIC#3446

Both the new Kong error format and KIC parsing are completely new, however, so I expect we may encounter things we want to change on either side. Does therefore it make sense to keep the go-kong body handling loose (i.e. it provides the body to downstream as-is and doesn't care about the contents at all itself) or lock that into go-kong now?

With the controller as the sole user of the new format for the time being, I think we may want to leave the body parsing logic (and associated version-gated tests) there for now, see if we get feedback from the field, and if the initial implementation seems solid, graduate body parsing into go-kong functionality? IMO we create an issue to collect any feedback on the KIC usage and follow-up on making it generic go-kong functionality later.

kong/config_service.go
```diff
@@ -37,29 +38,25 @@ func (c *ConfigService) ReloadDeclarativeRawConfig(
...
+	var b []byte
 	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
-		b, err := io.ReadAll(resp.Body)
+		b, err = io.ReadAll(resp.Body)
 		if err != nil {
-			return fmt.Errorf(
-				"failed posting new config to /config: got status code %d (and failed to read the response body): %w",
-				resp.StatusCode, err,
-			)
+			return []byte{},
```
pmalek, line 54: When I look at this now I believe returning a `nil` makes no sense. Why allocate an empty slice?

Actually: after looking at https://groups.google.com/g/golang-nuts/c/bqeemNtZkb4 it seems that there's little to no difference. I'd prefer to return `nil` to make it explicit though. WDYT?

kong/config_service.go
```diff
@@ -13,7 +12,8 @@ type AbstractConfigService interface {
 	// API endpoint using the provided reader which should contain the JSON
 	// serialized body that adheres to the configuration format specified at:
 	// https://docs.konghq.com/gateway/latest/production/deployment-topologies/db-less-and-declarative-config/#declarative-configuration-format
-	ReloadDeclarativeRawConfig(ctx context.Context, config io.Reader, checkHash bool) error
+	// It returns the response body and an error, if it encounters any.
+	ReloadDeclarativeRawConfig(ctx context.Context, config io.Reader, checkHash bool) ([]byte, error)
```
pmalek, line 16: This deserves a Changelog entry in my eyes.

pmalek (commented):
Just 2 nits.

As for processing/manipulating the body in go-kong vs KIC: I'm fine with leaving it here. It's actually better I'd say.

As for the tests: I just wanted to suggest a test where we assert for a given request/endpoint/payload tuple we get an expected body returned from `ReloadDeclarativeRawConfig`.

kong/config_service.go
rainest: No reason for the empty slice. I just don't naturally think in terms of acceptable Go conversions, so I see `[]byte` return signature, I return `[]byte{}` if I have nothing, since that's more clearly a `[]byte`.

I changed it and then read the thread and realized that it doesn't matter. But hey, it doesn't matter.

kong/config_service.go
rainest: Indeed. Also need to use this soon, so added one as a new release for tomorrow.

rainest:
golangci-lint apparently decided it should be upset about year-old code in this PR. I opened https://github.com/Kong/go-kong/issues/273 since IDK exactly what we're doing with that copy instead of read, and fixing it is out of scope for this PR.

kong/config_service.go
```diff
@@ -37,29 +38,25 @@ func (c *ConfigService) ReloadDeclarativeRawConfig(
...
+	var b []byte
 	if resp.StatusCode < 200 || resp.StatusCode >= 400 {
-		b, err := io.ReadAll(resp.Body)
+		b, err = io.ReadAll(resp.Body)
 		if err != nil {
-			return fmt.Errorf(
-				"failed posting new config to /config: got status code %d (and failed to read the response body): %w",
-				resp.StatusCode, err,
-			)
+			return nil,
```
pmalek, line 54: Now this is inconsistent with lines 46 and 41 🙃

kong/client.go
```diff
@@ -238,7 +238,8 @@ func (c *Client) DoRAW(ctx context.Context, req *http.Request) (*http.Response,
 func (c *Client) Do(ctx context.Context, req *http.Request,
 	v interface{},
 ) (*Response, error) {
-	resp, err := c.DoRAW(ctx, req)
+	// TODO https://github.com/Kong/go-kong/issues/273 clear this lint ignore
+	resp, err := c.DoRAW(ctx, req) //nolint:bodyclose
```
pmalek, line 242: Might have as well just added a `defer`ed `resp.Close()` 🤷

rainest: Okay, that _does_ break the tests _horribly_: https://github.com/Kong/go-kong/actions/runs/4079171550/jobs/7030257554

Somehow that response body is `nil` when the deferred close files? What the heck are we doing that causes that? I guess somehow the copy and decode destroys it? Anyway, back to the original plan.

kong/config_service.go
rainest: https://github.com/Kong/go-kong/pull/271/commits/6a88d74095232a710a6c08b7bc67aeab668e87e8

rainest:
https://github.com/Kong/go-kong/pull/271/commits/7008ad752510aefda4990703fd74886694c210bc makes the suggested change from chat discussion, which removes the config service and folds its functionality into the Kong client.

rainest:
f7b64c0cc33fd92646133257ac035075091f62a5 because I forgot to move the test after removing the config service.

ab5b18a since something I did as part of that broke error reporting, and because the old error semantics were a bit off when always returning the body.

rainest:
Alright, one more for the road. Added an additional breaking feature, a61b0d5, because https://github.com/Kong/kong/pull/10161 requires a new query string to return the error format that https://github.com/Kong/kubernetes-ingress-controller/pull/3446 wants to use.

pmalek:
I made some minor tweaks to the test itself and added the body contents to the error we return in case the response code indicates an error.

But there's actually a more important thing here: the test fails with the following error:

```
{"name":"invalid declarative configuration","message":"declarative config is invalid: {flatten_errors=\"unknown field\"}","fields":{"flatten_errors":"unknown field"},"code":14}
```

The reason for this is that we don't test against a dbless kong 😱 https://github.com/Kong/go-kong/blob/4b75690bd2f834998cbbcb70d59456dc270672a0/.ci/setup_kong.sh#L49

kong/client.go
pmalek: The problem here is that we have an API: [`Client.Do()`](https://github.com/Kong/go-kong/blob/b5e8261e4507d18ab51d25c740152ada25466963/kong/client.go#L238) where we return a wrapper around `http.Response` where we don't close the body but it's already read (through `json.NewDecoder(resp.Body).Decode(v)`).

We should explicitly add a comment to that method that the response can be used to read e.g. the status code but the body will be completely read **BUT** the caller should close it nonetheless. Otherwise we might leak that.

Created #276 to track this.

pmalek: `DoRAW()` seem to be fine because it explicitly states in the comment that 

```
the caller is responsible for closing the response body
```

and body is not manipulated in anyway.

rainest:
> I made some minor tweaks to the test itself and added the body contents to the error we return in case the response code indicates an error.

Do we indeed want to do this? I'd actually removed this from the original error since we're now always returning bodies unless we can't read them, so if you get an error back you have the body, and can either log it or inspect it as needed. That body could be quite large, so I figured it made sense to omit it from the error string if you don't care beyond a brief explanation.

> But there's actually a more important thing here: the test fails with the following error

This is... odd. Is it somehow interpreting the query string as a field in some cases? Doing a packet capture shows no `flatten_errors` in the body, and attempting to replicate the request manually succeeds:

https://gist.github.com/rainest/b956bf4db29abb04418fd002bf3b0a1a

That looks like a gateway-side bug maybe.

rainest:
Alright, https://github.com/Kong/go-kong/pull/271/commits/348dd0881afdcf75753dfefde955c3749e3f0e40 to fix the error. Kong has an odd "bug" (not sure, since we appear to be aware of the behavior--we code around it--but keep it as-is) where query string fields are treated as part of the config body. Special-case known fields (`flatten_errors` and `check_hash`) are explicitly stripped by the API code, but versions prior to 3.2 don't strip `flatten_errors` because they don't know about it.

I have no idea why we do this, but the upshot is we need version-dependent behavior for part of the test only. This isn't something go-kong typically does, so the change adds some borrowed code from the "skip entire test by version" helper.

flrgh:
> This is... odd. Is it somehow interpreting the query string as a field in some cases? Doing a packet capture shows no `flatten_errors` in the body, and attempting to replicate the request manually succeeds:


For some more background, this is an unfortunate quirk of the [web framework that the admin API uses](https://leafo.net/lapis/reference/actions.html#request-object). It lumps query args, form parameters, and I believe even request body JSON all into one single object (`self.params`) that is passed to the endpoint/route handler:

| Name | Description |
|------|-------------|
| self.params | A table containing all request parameters merged together, including query parameters and form-encoded parameters from the body of the request. See [Request Parameters](https://leafo.net/lapis/reference/actions.html#request_parameters) for more details. |

So `POST /config?foo=bar` with a JSON content-type and JSON body of `{ "baz": "bat" }` yields a lua table that looks like `{ foo = "bar", baz = "bat" }` for `self.params`. We process this table by parsing and stripping known query args, but any _unknown_ query args will wind up being passed through to the declarative config schema validator.

kong/client_test.go
```diff
@@ -205,3 +207,123 @@ func TestBaseRootURL(t *testing.T) {
...
+	// because it thinks "flatten_errors" is an actual (and invalid) field inside the config.
+	//
+	// This is the one test where we want version-dependent behavior, but only for one value within the test. We test
+	// config updates on all DB-less capable versions, but only set flattenErrors=true on 3.2+. To handle that, this
+	// snippet is borrowed from RunWhenKong, to allow toggling that behavior only on or off depending on the version.
+	var flattenErrors bool
+	client, err := NewTestClient(nil, nil)
+	if err != nil {
+		t.Error(err)
+	}
```
pmalek, line 279: Why no `require.NoError(t, err)`? Much less boilerplate

Ditto elsewhere

kong/client_test.go
```diff
@@ -205,3 +207,123 @@ func TestBaseRootURL(t *testing.T) {
...
+	r, err := NewRange(">=3.2.0")
+	if err != nil {
+		t.Error(err)
+	}
+	if r(currentVersion) {
+		t.Log("Kong version >=3.2, enabling flattenErrors for ReloadDeclarativeRawConfig")
+		flattenErrors = true
+	} else {
+		t.Log("Kong version <3.2, disabling flattenErrors for ReloadDeclarativeRawConfig")
+	}
```
pmalek, line 298: Just a thought that instead of doing all these shenanigans we could just add a `func()` field in table driven test struct which will call appropriate `RunWhenKong(T, ">=3.2")` etc making that version filtering be bound to a concrete test case e.g.

When done this way a potential test failure will be much more specific (subjectively ?)

kong/client_test.go
rainest: Literally just copy-pasta from the source. No idea why the helper's written that way (we do generally want to avoid `require`s in helpers, but that's a more broad advice against avoiding any interaction with the test struct in helpers, and we were already doing that with the `t.Error(err)`s.

kong/client_test.go
rainest: Not quite sure I follow, can you show a code example or add a commit?

The other thing I'd considered was making the go-kong test helper suite support a second set of lighter `IfKongVersion(">=3.2")` helpers that did not interact with the test object directly (hey! same advice from the other comment that we aren't following yet :laughing: ), but I figured that, if I was going to build that out, it should have support across the board for the current whole-test helpers, so `IfKongEnterprise(...)` also and so on. So long as this is the odd test out, I YAGNIed on that and instead brought the version checking logic under this one oddball test.

kong/client_test.go
```diff
@@ -205,3 +207,116 @@ func TestBaseRootURL(t *testing.T) {
...
+				assert.Errorf(t, err, "Client.SendConfig() got unexpected error = %v", err)
+			} else {
+				assert.NoError(t, err)
+			}
+
+			// this is somewhat untrue: network or HTTP-level failures _can_ result in a nil response body. however,
+			// none of our test cases should cause network or HTTP-level failures, so fail if they do occur. if this
+			// _does_ encounter such a failure, we need to investigate and either update tests or fix some upstream bug
+			// if it's not some transient issue with the testing environment
+			require.NotNilf(t, body, "body was nil; should never be nil")
```
pmalek, line 319: With this in place we can remove the assertion above that checks the body

kong/client_test.go
```diff
@@ -205,3 +207,116 @@ func TestBaseRootURL(t *testing.T) {
...
+			b, err := json.Marshal(tt.config)
+			require.NoError(t, err)
+
+			body, err := client.ReloadDeclarativeRawConfig(ctx, bytes.NewBuffer(b), true, flattenErrors)
+			// We only get empty body when there's a transient network error or
+			// we fail to read the response body which shouldn't happen in tests.
+			assert.NotEmpty(t, string(body))
+
+			if tt.wantErr {
+				assert.Errorf(t, err, "Client.SendConfig() got unexpected error = %v", err)
```
pmalek, line 310: Assert package will pretty print the error in case it's not nil. You might as well drop it from the format string

pmalek (approved):

PR #275, opened 2023-02-03
ci: add testing against dbless

pmalek:
This also bumps default kong image to 3.1.1 and makes some minor refactor in test skipping conditionals (marking `t.Helper()` to prevent logging helper line numbers but to print corresponding test line numbers)

### DIFF
--- a/.ci/setup_kong.sh
+++ b/.ci/setup_kong.sh
@@ -93,7 +93,7 @@ function deploy_kong_dbless()
 
 function create_network()
 {
-  docker network inspect $NETWORK_NAME >/dev/null|| docker network create $NETWORK_NAME
+  docker network inspect $NETWORK_NAME 2>/dev/null >/dev/null || docker network create $NETWORK_NAME
 }
 
 while [[ $# -gt 0 ]]; do
@@ -113,8 +113,6 @@ while [[ $# -gt 0 ]]; do
     *)
   esac
 done
-
-set -x
 
 if [[ "${DBMODE}" == "off" ]]; then
   create_network
--- a/.ci/setup_kong.sh
+++ b/.ci/setup_kong.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-KONG_IMAGE=${KONG_IMAGE_REPO:-kong}:${KONG_IMAGE_TAG:-3.0.0}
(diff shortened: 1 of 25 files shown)


rainest (approved):
Do we maybe want to default to always testing by default only in Postgres, and then having the DB-less tests a special case where we mark them DB-less and they then only run in DB-less, or maybe both?

Most of the go-kong tests require adding resources and then getting the same resources, and that pattern only works in DB-backed mode. We could rework them to have an alternate GET-only path that relies on a static config, but it doesn't look like we're doing that here (it would require a fairly large test refactor).

There are some outliers like status endpoint that run the same in either, but AFAIK they're by far in the minority.

The `/config` functionality is the main bit that really needs DB-less, and doesn't work at all in Postgres.

Given that the service tests require Postgres and are the majority, assuming `RunWhenDBMode(T, "postgres")` if we don't call it explicitly with another value would simplify new tests a bit, so you don't need to remember.

That said, test writing here tends to involve some degree of copy/paste, so approving anyway given that it's not that huge a burden. Leaving open in case any of the other requested reviewers want to poke at it.

pmalek:
> Given that the service tests require Postgres and are the majority, assuming `RunWhenDBMode(T, "postgres")` if we don't call it explicitly with another value would simplify new tests a bit, so you don't need to remember.

I understand the concern about copy/paste but OTOH better this way than get hit by a failure because of an implicit default (e.g. testing for postgres only by default).

I don't know the specifics of tests here enough to comment but I do find what is proposed in this PR as fairly straightforward setup that reads well.

rainest:
Alright, https://github.com/Kong/go-kong/pull/271/commits/348dd0881afdcf75753dfefde955c3749e3f0e40 to fix the error. Kong has an odd "bug" (not sure, since we appear to be aware of the behavior--we code around it--but keep it as-is) where query string fields are treated as part of the config body. Special-case known fields (`flatten_errors` and `check_hash`) are explicitly stripped by the API code, but versions prior to 3.2 don't strip `flatten_errors` because they don't know about it.

I have no idea why we do this, but the upshot is we need version-dependent behavior for part of the test only. This isn't something go-kong typically does, so the change adds some borrowed code from the "skip entire test by version" helper.

pmalek:
I'm going ahead an merge this so that we can keep working on this in iterations.

I'd personally prefer not to make this "test on postgres only by default" like it was :)

PR #277, opened 2023-02-07
ci(lint): setup YAML Lint step

mmorel-35:
Defines a step to lint YAML files with yamllint

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,12 @@
 version: 2
 updates:
-- package-ecosystem: gomod
-  directory: /
-  target-branch: main
-  schedule:
-    interval: daily
-- package-ecosystem: github-actions
-  directory: /
-  target-branch: main
-  schedule:
-    interval: daily
+  - package-ecosystem: gomod
+    directory: /
+    target-branch: main
+    schedule:
+      interval: daily
+  - package-ecosystem: github-actions
+    directory: /
+    target-branch: main
+    schedule:
+      interval: daily
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,10 +13,10 @@ name: "CodeQL"
 
 on:
   push:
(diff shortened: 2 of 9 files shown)


pmalek:
Hi 👋 

Thanks for this contribution. Before we move on let us fix the CI pipeline for outside contributors (where we don't have access to secrets) #278.

In the meantime: we're trying to use the same set of tools, versions and configurations across CI and local environments. This has worked pretty well for us so far.

You can find how that's defined e.g. in KIC's repo in https://github.com/Kong/kubernetes-ingress-controller/blob/36b520680df37cb1e04d9504e1fc2f93eceea83b/Makefile#L33-L74 along with a separate module for third party tools that we use and version separately https://github.com/Kong/kubernetes-ingress-controller/tree/main/third_party.

If you'd like to continue with this contribution I suggest that we take this approach and appropriately add a Makefile target and a corresponding set of files as we do in KIC.

WDYT?

rainest:
Anything in particular this is trying to fix? YAML is, for better or worse, flexible with regards to indents and whatever in some contexts. While consistency is good, it's hard to guarantee that local environments and projects will use the same linters or linter settings. The evergreen "array items should match parent indent" or "array items should be indented one level beyond parent" debate ends up being an annoying bikeshed-y thing IMO: YAML treats either as valid, and virtually nobody seems to agree on which is better.

I would avoid enforcing it as an unnecessary point of friction: so long as the YAML is valid, it's acceptable. If it matters that much, the YAML standard itself needs to change to require one or the other. One-off updates to make like files use consistent formatting are okay, but we shouldn't require one or the other unless there's a broader community standard (e.g. if Kubernetes decides that it wants to prefer one style, we should probably follow it).

A lot of this is handled via standardization in generators (AFAIK Kustomize and the K8S API server produce consistent YAML by converting to JSON and back). Edge stuff like GH Actions configuration I'd say just let be.

PR #281, opened 2023-02-10
refactor: organizing types

tjasko:
This change simply reorganizes the existing types into their own files. Other than moving types to their own file, no other code was updated in the first commit.

This work is being done as there are plans to introduce more methods defined on each type, e.g.: sorting methods, so that we are able to sort an entire Kong config payload (which is useful for deterministic comparisons, like in tests). Without this change, the `types.go` file would get a bit out of control.


kong/sni.go
```diff
@@ -0,0 +1,23 @@
...
+// +k8s:deepcopy-gen=true
+type SNI struct {
+	ID          *string      `json:"id,omitempty" yaml:"id,omitempty"`
+	Name        *string      `json:"name,omitempty" yaml:"name,omitempty"`
+	CreatedAt   *int64       `json:"created_at,omitempty" yaml:"created_at,omitempty"`
+	Certificate *Certificate `json:"certificate,omitempty" yaml:"certificate,omitempty"`
+	Tags        []*string    `json:"tags,omitempty" yaml:"tags,omitempty"`
+}
+
+// FriendlyName returns the endpoint key name or ID.
```
pmalek, line 14: I believe this comment is copy pasted to each type not necessarily describing each of the respective `FriendlyName()`s.

Is "returns the endpoint key name or ID" accurate for e.g. SNI? 🤔

tjasko: This is all existing code (as indicated by the commit/PR description), the only thing I actually changed was this:
https://github.com/Kong/go-kong/pull/281/commits/e2a20300f0a2c55c84621786f0f512db68972add

We can review these things if we want, but I wanted to keep this PR as simple as possible given how much refactoring it's already doing.

pmalek: Fair enough. Let's leave it as is and tackle those issues separately. 👍

tjasko: Great, thanks @pmalek. 👍

pmalek (approved):

PR #300, opened 2023-03-27
feat: add global query params to kong client

aboudreault:
This is mostly a konnect requirement for our development workflow, but there are cases that we need to set global query params to use deck+go-kong.

### DIFF
--- a/kong/client.go
+++ b/kong/client.go
@@ -34,8 +34,10 @@ var defaultCtx = context.Background()
 // Client talks to the Admin API or control plane of a
 // Kong cluster
 type Client struct {
-	client                  *http.Client
-	baseRootURL             string
+	client      *http.Client
+	baseRootURL string
+	// QueryParams are query string parameters that are set on all requests.
+	QueryParams             url.Values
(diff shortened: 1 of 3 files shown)


rainest (commented):
What's the actual context here? "Because our workflow needs it" without further explanation or links to supporting design docs is quite vague, and global query parameters that should apply to all endpoints seem a bit odd.

aboudreault:
I will provide a link with more context asap (which is private).

mflendrich:
This PR appears to be introduction an extension that is outside of the Admin API protocol which can make maintenance of the Admin API surface problematic. Pinging @hbagdi @mikefero @dndx to agree together on how to solve for such differences in unrelated Admin API implementations.

I think that we should strive towards having a strict standard for an Admin API without weakly defined customizations incompatible across implementations.

aboudreault:
let's put this on hold until we check the alternatives. Some options:
1- to avoid this change, implement the support to accept a header instead of a query param (we will check this one first)
2- provide an new client utility function similar to https://github.com/Kong/go-kong/blob/main/kong/utils.go#L105

PR #304, opened 2023-04-03
feat: add license endpoints

rainest:
Adds and tests the `/licenses` endpoint for Enterprise. Releases 0.40.

The test requires a `KONG_LICENSE_DATA` environment variable with a valid license. AFAIK we have no designated testing license that we can commit to a public codebase, so this relies on the CI license.

The test does override the initial license after the test suite starts. There's not really any way to isolate that in the go-kong test suite, but it does not appear to cause any issues. It's overriding to the same license it uses initially anyway.


kong/license_service_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLicensesService(t *testing.T) {
+	RunWhenEnterprise(t, ">=2.3.0", RequiredFeatures{})
+	require := require.New(t)
+	assert := assert.New(t)
+	licenseString := os.Getenv("KONG_LICENSE_DATA")
```
pmalek, line 16: This test as far as I understand doesn't make sense when `KONG_LICENSE_DATA` is empty right? Maybe we 

- check that here and skip the test when that's the case
- add a similar helper func which would skip the test if the license is not precent, similar to `RunWhenEnterprise()`

WDYT?

rainest: I think the Enterprise tests fail fast and run nothing when they have no license, but will check. That's probably CI-only though, so may as well add a code-level check.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Table of Contents
 
+<<<<<<< HEAD
+=======
```
pmalek, line 4: Conflict resolution went awry

CHANGELOG.md
```diff
@@ -727,7 +737,11 @@ authentication credentials in Kong.
   releases of Kong since every release of Kong is introducing breaking changes
   to the Admin API.
 
+<<<<<<< HEAD
 [v0.39.2]: https://github.com/Kong/go-kong/compare/v0.39.1...v0.39.2
+=======
+[v0.40.0]: https://github.com/Kong/go-kong/compare/v0.39.1...v0.40.0
+>>>>>>> 59e462b (feat: support license endpoint)
```
pmalek, line 744: ditto

CHANGELOG.md
```diff
@@ -51,10 +52,17 @@
 - [0.2.0](#020)
 - [0.1.0](#010)
 
-## [v0.39.2]
+## [v0.40.0]
 
 > Release date: 2023/04/06
```
rainest, line 57:
```suggestion
> Release date: 2023/04/07
```

pmalek (approved):

Issue #307, opened 2023-04-05
Record default filler removes populated fields


rainest:
https://github.com/Kong/go-kong/pull/302 introduced a bug that was not visible until integrated into another application.

[bug.tar.gz](https://github.com/Kong/go-kong/files/11162123/bug.tar.gz) shows a demonstration using [deck](https://github.com/Kong/deck). Syncing `orig.yaml` with the current deck release (`/tmp/deck`) and then diffing with a build that includes 0.39.1 shows it removing the populated `config.add.headers` array.

nishant95:
@rainest Sure, i'll look into it.

nishant95:
raised a PR #309

nishant95:
@rainest raised a PR #309 with the fix. 

Can we add some checks so that such bugs can be detected before merging the PR?

rainest:
We can't easily add consistent checks beyond the style of tests already in this repo since it'd mean running other repo's tests.

I should have run a manual replace test like https://github.com/Kong/kubernetes-ingress-controller/pull/3866 earlier.

nishant95:
Fixed in #309

PR #309, opened 2023-04-06
Fix: fill config defaults

nishant95:
This has the fix applied in #302 along with resolution for  #307 

Added test for the bug encountered in #307 as well.


kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
 		})
 	}
 }
+
+func Test_fillConfigRecord(T *testing.T) {
```
pmalek, line 1004:
```suggestion
func Test_fillConfigRecord(t *testing.T) {
```

to prevent confusion between and accidentally failing the whole test where only a subtest should fail.

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		T.Run(tc.name, func(t *testing.T) {
+			configSchema, err := getConfigSchema(tc.schema)
+			assert.NoError(err)
+			config := fillConfigRecord(configSchema, tc.config)
+			assert.NotNil(config)
```
pmalek, line 1073:
```suggestion
		t.Run(tc.name, func(t *testing.T) {
			configSchema, err := getConfigSchema(tc.schema)
			require.NoError(t, err)
			config := fillConfigRecord(configSchema, tc.config)
			require.NotNil(t, config)
```

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+			if diff := cmp.Diff(config, tc.expected); diff != "" {
+				t.Errorf(diff)
+			}
+		})
+	}
+}
+
+func Test_FillPluginsDefaults(T *testing.T) {
```
pmalek, line 1081:
```suggestion
func Test_FillPluginsDefaults(t *testing.T) {
```

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+	RunWhenKong(T, ">=2.6.0 <2.8.1")
+	assert := assert.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 1087:
```suggestion
	client, err := NewTestClient(nil, nil)
	require.NoError(t, err)
	require.NotNil(t, client)
```

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		T.Run(tc.name, func(t *testing.T) {
+			plugin := tc.plugin
+			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/statsd")
+			assert.NoError(err)
+			assert.NotNil(fullSchema)
```
pmalek, line 1129:
```suggestion
		t.Run(tc.name, func(t *testing.T) {
			plugin := tc.plugin
			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/statsd")
			require.NoError(t, err)
			require.NotNil(t, fullSchema)
```

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+		})
+	}
+}
+
+func Test_FillPluginsDefaults_RequestTransformer(T *testing.T) {
+	assert := assert.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 1148:
```suggestion
func Test_FillPluginsDefaults_RequestTransformer(t *testing.T) {
	client, err := NewTestClient(nil, nil)
	require.NoError(t, err)
	require.NotNil(t, client)
```

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		T.Run(tc.name, func(t *testing.T) {
+			plugin := tc.plugin
+			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/request-transformer")
+			assert.NoError(err)
+			assert.NotNil(fullSchema)
```
pmalek, line 1230:
```suggestion
		t.Run(tc.name, func(t *testing.T) {
			plugin := tc.plugin
			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/request-transformer")
			require.NoError(t, err)
			require.NotNil(t, fullSchema)
```

kong/utils_test.go
```diff
@@ -999,3 +1000,241 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
+			assert.NotNil(config)
+			if diff := cmp.Diff(config, tc.expected); diff != "" {
+				t.Errorf(diff)
+			}
+		})
+	}
+}
+
+func Test_FillPluginsDefaults(T *testing.T) {
+	RunWhenKong(T, ">=2.6.0 <2.8.1")
```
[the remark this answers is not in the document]
nishant95: @pmalek 
We wanted to add tests for a default plugin which was impacted by this bug (discussed [here](https://github.com/Kong/go-kong/pull/302#pullrequestreview-1363700741)).

But the schema for this plugin has changed a lot in the newer versions (and we are fetching the schema from Kong in this test), hence I added this version check so that it doesn't fail unnecessarily.
Probably we can handle it differently so that the test is valid for all versions of kong.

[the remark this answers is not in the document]
pmalek: Got it. Perhaps we could add another test that would work with more recent versions of Kong? We can have those either as separate tests named accordingly or as subtests guarded by the `RunWhenKong()` with appropriate versions range. WDYT?

If we leave it like that what would ensure us that users working with plugins in Gateway 3.2 are getting the correct behavior?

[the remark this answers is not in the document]
nishant95: We should definitely add tests for the recent versions. 

One way to do this is to add schema json in the test data itself, instead of fetching it from Kong. But looking at a few other tests it seems we are no longer doing that (Any reason for this?). 

Also, this particular test scenario might get removed if the StatsD plugin changes its schema to remove the array of records from its config in newer versions.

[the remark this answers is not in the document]
pmalek: You're right. In that case: would be a reasonable ask to suggest that we don't rely on the client itself and just test `FillPluginsDefaults` by specifying the schema in the test? This way we're not relying on the Gateway version and we're only concerned about the function that we're testing, not the environment.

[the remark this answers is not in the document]
nishant95: @pmalek I've added schema to the test.

Also, while adding a few more tests I discovered that default values were not being applied for `nil` array of records (working fine for empty array). Added handling for this as well.

kong/utils_test.go
```diff
@@ -999,3 +1203,310 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
 		})
 	}
 }
+
+func Test_fillConfigRecord(t *testing.T) {
+	client, err := NewTestClient(nil, nil)
+	require.NoError(t, err)
+	require.NotNil(t, client)
+
```
pmalek, line 1211: We don't need this anymore :) 

```suggestion

```

kong/utils_test.go
```diff
@@ -12,8 +12,212 @@ import (
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/tidwall/gjson"
 )
 
+var StatsDSchema = `{
```
pmalek, line 18: I believe this can be made `const`

```suggestion
const StatsDSchema = `{
```

pmalek (commented):
Thanks! 👍 Some minor nits.

I'll let @rainest double check when he comes online but looks good to me.

rainest (approved):
https://github.com/Kong/kubernetes-ingress-controller/pull/3866 trial run integrated into KIC is passing Knative and conformance now, so extra confirmation in integration that the Request Transformer configuration is no longer getting clobbered.

https://github.com/Kong/deck/pull/888 now succeeds as well, where it was failing with 0.39.1

nishant95:
this fixes #307

aboudreault:
@nishant95 @pmalek @rainest looks like this change introduced another bug where an array is completely ignored. we can reproduce with deck and the following:
```
_format_version: "3.0"

plugins:
- name: kafka-log
  enabled: false
  config:
    topic: konnect.auditlogs.gateway-access-logs
    bootstrap_servers:
    - host: broker-1.endpoint
      port: 9096
    - host: broker-1.endpoint
      port: 9096
    - host: broker-1.endpoint
      port: 9096
    authentication:
      strategy: sasl
      mechanism: SCRAM-SHA-512
      user: sasl_user
      password: xxxxxxxx
    security:
      ssl: false
    cluster_name: konnect-msk
```

aboudreault:
ref: https://github.com/Kong/go-kong/pull/333

nishant95:
@aboudreault #309  only adds handling for subconfig of type "array", how does this cause a subconfig of type "set" to fail?
How was this working before this PR was merged?

Also, the [handle_missing_fields](https://github.com/Kong/kong/blob/241c4d0681c7996e8565e278937838f5ec723273/kong/db/schema/init.lua#L1001) in lua code and [FillPluginsDefaults](https://github.com/Kong/go-kong/blob/3ff328d54ff853d7bf24e5f1ab04adb8fd22a317/kong/utils.go#L510) in go-kong seems out of sync. We recently faced an issue in case of non-required subconfig of type record (Testing the fix, yet to raise a PR). 
Kong kept the default as nil but KIC filled the default with a record with nil fields causing KIC to be stuck in sync retry loop and ultimately getting OOMKilled.

nishant95:
@pmalek @rainest @aboudreault 
Fix for #334: #336

PR #316, opened 2023-04-17
chore(tests) split out plugin map test

rainest:
Separate the plugin arbitrary map default test from the main default test. The new test only checks the arbitrary map field.

Changes to the test plugin in upstream Kong break the blanket diff used in the main test, and would require a version split without testing only the relevant field.

The issue appeared in https://github.com/Kong/go-kong/actions/runs/4718806033/jobs/8369120871?pr=315

```
 === RUN   TestFillPluginDefaults/nested_config_with_arbitrary_map_field
    plugin_service_test.go:706:   &kong.Plugin{
          	... // 5 identical fields
          	Service:  nil,
          	Consumer: nil,
          	Config: kong.Configuration{
          		"content_type":         string("application/json"),
          		"custom_fields_by_lua": map[string]any{"foo": string("bar")},
        - 		"flush_timeout":        nil,
        + 		"flush_timeout":        float64(2),
          		"headers":              nil,
          		"http_endpoint":        nil,
          		"keepalive":            float64(60000),
          		"method":               string("POST"),
        - 		"queue": map[string]any{
        - 			"initial_retry_delay":  float64(0.01),
        - 			"max_batch_size":       float64(1),
        - 			"max_bytes":            nil,
        - 			"max_coalescing_delay": float64(1),
        - 			"max_entries":          float64(10000),
        - 			"max_retry_delay":      float64(60),
        - 			"max_retry_time":       float64(60),
        - 			"name":                 nil,
        - 		},
        - 		"queue_size":  nil,
        + 		"queue_size":  float64(1),
        - 		"retry_count": nil,
        + 		"retry_count": float64(10),
          		"timeout":     float64(10000),
          	},
          	Enabled: &false,
          	RunOn:   nil,
          	... // 3 identical fields
          }
--- FAIL: TestFillPluginDefaults (0.04s)
```

These fields were added upstream for 3.3: https://github.com/Kong/kong/commit/9df893f6aff98cd51f27f1c27fa30fdcf13fcf48

### DIFF
--- a/kong/plugin_service_test.go
+++ b/kong/plugin_service_test.go
@@ -658,6 +658,38 @@ func TestFillPluginDefaults(T *testing.T) {
 				Enabled:   Bool(false),
 			},
 		},
+	}
+
+	for _, tc := range tests {
+		T.Run(tc.name, func(t *testing.T) {
+			p := tc.plugin
+			fullSchema, err := client.Plugins.GetFullSchema(defaultCtx, p.Name)
+			assert.NoError(err)
+			assert.NotNil(fullSchema)
+			if err := FillPluginsDefaults(p, fullSchema); err != nil {
+				t.Errorf(err.Error())
+			}
+
+			if diff := cmp.Diff(p, tc.expected); diff != "" {
+				t.Errorf(diff)
+			}
+		})
+	}
+}
+
+// TestFillPluginDefaultsArbitraryMap is split from TestFillPluginDefaults due to a version compatibility issue
+func TestFillPluginDefaultsArbitraryMap(T *testing.T) {
+	RunWhenKong(T, ">=2.3.0")
+
+	client, err := NewTestClient(nil, nil)
+	require.NoError(T, err)
+	require.NotNil(T, client)
+
+	tests := []struct {
+		name     string
+		plugin   *Plugin
+		expected *Plugin
+	}{
 		{
 			name: "nested config with arbitrary map field",
 			plugin: &Plugin{
@@ -696,13 +728,16 @@ func TestFillPluginDefaults(T *testing.T) {
 		T.Run(tc.name, func(t *testing.T) {
 			p := tc.plugin
 			fullSchema, err := client.Plugins.GetFullSchema(defaultCtx, p.Name)
-			assert.NoError(err)
-			assert.NotNil(fullSchema)
+			require.NoError(t, err)
+			require.NotNil(t, fullSchema)
 			if err := FillPluginsDefaults(p, fullSchema); err != nil {
 				t.Errorf(err.Error())
 			}
 
-			if diff := cmp.Diff(p, tc.expected); diff != "" {
+			// the log plugins are the only plugins that use the typedefs.lua_code type in their schema
+			// https://github.com/Kong/kong/commit/9df893f6aff98cd51f27f1c27fa30fdcf13fcf48 changes a number of other
+			// fields for 3.3, so this test only checks the relevant field to avoid needing a version split
+			if diff := cmp.Diff(p.Config["custom_fields_by_lua"], tc.expected.Config["custom_fields_by_lua"]); diff != "" {
 				t.Errorf(diff)
 			}
 		})
--- a/kong/plugin_service_test.go
+++ b/kong/plugin_service_test.go
@@ -550,11 +550,10 @@ func TestFillPluginDefaults(T *testing.T) {
 	// not all Enterprise versions.
 	SkipWhenEnterprise(T)
 	RunWhenKong(T, ">=2.3.0")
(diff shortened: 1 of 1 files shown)


kong/plugin_service_test.go
```diff
@@ -658,6 +658,39 @@ func TestFillPluginDefaults(T *testing.T) {
...
+
+	for _, tc := range tests {
+		T.Run(tc.name, func(t *testing.T) {
+			p := tc.plugin
+			fullSchema, err := client.Plugins.GetFullSchema(defaultCtx, p.Name)
+			assert.NoError(err)
+			assert.NotNil(fullSchema)
+			if err := FillPluginsDefaults(p, fullSchema); err != nil {
+				t.Errorf(err.Error())
+			}
```
pmalek, line 671: That's an unfortunate issue with several tests that I've seen around here:

- using `assert` where we should use `require`
- using `assert` object based on the supertest's object: this will fail the whole test, while it should only fail the subtest

```suggestion
			require.NoError(t, err)
			require.NotNil(t, fullSchema)
			require.NoError(t, FillPluginsDefaults(p, fullSchema))
```

kong/plugin_service_test.go
```diff
@@ -658,6 +658,39 @@ func TestFillPluginDefaults(T *testing.T) {
...
+}
+
+// TestFillPluginDefaultsArbitraryMap is split from TestFillPluginDefaults due to a version compatibility issue
+func TestFillPluginDefaultsArbitraryMap(T *testing.T) {
+	RunWhenKong(T, ">=2.3.0")
+	assert := assert.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 687: ditto

```suggestion
	require.NoError(t, err)
	require.NotNil(t, client)
```

pmalek: + ditto in lines below

kong/plugin_service_test.go
```diff
@@ -702,7 +735,10 @@ func TestFillPluginDefaults(T *testing.T) {
 				t.Errorf(err.Error())
 			}
 
-			if diff := cmp.Diff(p, tc.expected); diff != "" {
+			// the log plugins are the only plugins that use the typedefs.lua_code type in their schema
+			// https://github.com/Kong/kong/commit/9df893f6aff98cd51f27f1c27fa30fdcf13fcf48 changes a number of other
+			// fields for 3.3, so this test only checks the relevant field to avoid needing a version split
+			if diff := cmp.Diff(p.Config["custom_fields_by_lua"], tc.expected.Config["custom_fields_by_lua"]); diff != "" {
```
pmalek, line 741: I wonder if it would be better to actually do this version split. BTW how would you see this exactly? 🤔

pmalek (changes requested):

kong/plugin_service_test.go
rainest: See what, the failure? It popped up on the unreleased Kong test on some random Dependabot PR, and the diff is basically "the defaults for this plugin changed because its schema changed".

I don't think we'd want the version split, since we'd need to maintain two copies of the test. The field the test cares about _doesn't_ change between versions, and we don't really care about the others. This case only cares about fields with this unusual type.

The original test test uses a diff across the entire output as a simple means of handling default tests without enumerating each field of interest for the individual cases. Most of the plugins have fairly static schemas, so that shortcut doesn't break across versions.

kong/plugin_service_test.go
pmalek: Why resolved but not comment and/or changed 😢 ?

kong/plugin_service_test.go
pmalek: Alright so maybe it's just me not immediately understanding the issue here.

After taking a look at it again, it's not super clear what we're trying to test here. I get the schema change in (yet to be released) 3.3 but I don't get the goal of this test.

We're checkout `p.Config["custom_fields_by_lua"]` but not the whole `p` or `p.Config` because? 

Also:

> so this test only checks the relevant field

**Why** is only that field relevant? The test name or comments do not say why it is. 

`FillPluginDefaults()` name implies that we're filling the plugin defaults, it doesn't say anything about being specific for **a field**.

The test (or the carved out piece of it) makes it hard to figure out for a reader without a context (e.g. knowing the implementation details of `FillPluginDefaults()`).

That makes sense?

pmalek (approved):

pmalek:
I've pushed the change to use require and assert based on correct (subtest's) `*testing.T`.

czeslavo (approved):

czeslavo:
I allow myself to merge this - it unblocks PRs to main.

PR #321, opened 2023-04-27
fix: allow explicit nulls in config

### LINKED ISSUE #320: Preserve explicit nulls in plugin configs
## Problem

Changes in [#309](https://github.com/Kong/go-kong/pull/309) filled default values into fields of plugin configurations that are explicitly set to `null`. Some plugin supports `null`s in certain fields of their configurations, for example `datadog` plugin allows `null` in `host` to load host from environment variable. 

## Proposed solution
- do not fill in default value if a field is explicitly set to `nil`. (may be switched by params)

## Related issue:
-  https://github.com/Kong/kubernetes-ingress-controller/issues/3908

randmonkey:
Allow explicit `null`s in configurations to fix #320 .

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -247,6 +247,10 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 				if v != nil {
 					return true
 				}
+				ftype := value.Get(fname + ".type").String()
+				if ftype != "record" && ftype != "array" {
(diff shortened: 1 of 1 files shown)


mflendrich:
Not planned - as explained in https://github.com/Kong/go-kong/issues/320#issuecomment-1812962666

PR #336, opened 2023-06-05
fix: fill defaults for non-required field of type "record"

### LINKED ISSUE #334: wrong default filled for non-required field of type "record"
### Problem

 [handle_missing_fields](https://github.com/Kong/kong/blob/241c4d0681c7996e8565e278937838f5ec723273/kong/db/schema/init.lua#L1001) in lua code and [FillPluginsDefaults](https://github.com/Kong/go-kong/blob/3ff328d54ff853d7bf24e5f1ab04adb8fd22a317/kong/utils.go#L510) in go-kong has different handling for default values in the following scenario:
- when the field is of type "record"
- `required` is set to false
- No subconfig provided for that field. 

Kong Gateway fills the default as `nil` but KIC fills the default with a record object with nil fields.
This causes KIC to keep detecting diff and it gets stuck in sync retry loop and ultimately gets OOMKilled.

nishant95:
Fixes #334 

Adds a check for required while recusrively filling defaults for subconfig of type record.
Current behaviour of KIC is out of sync from Kong Lua implementation.

There are some field "required" checks in Lua code: [here](https://github.com/Kong/kong/blob/2a3e9193c37a8df3eb3edae6728fa84b4bfed820/kong/db/schema/init.lua#L1020) and [here](https://github.com/Kong/kong/blob/2a3e9193c37a8df3eb3edae6728fa84b4bfed820/kong/db/schema/init.lua#L1538)
These checks are missing form go code

### DIFF
--- a/kong/utils_test.go
+++ b/kong/utils_test.go
@@ -1511,6 +1511,7 @@ func Test_FillPluginsDefaults_RequestTransformer(t *testing.T) {
 }
 
 func Test_FillPluginsDefaults_Acme(t *testing.T) {
+	RunWhenKong(t, ">=2.6.0")
 	client, err := NewTestClient(nil, nil)
 	require.NoError(t, err)
 	require.NotNil(t, client)
@@ -1592,7 +1593,7 @@ func Test_FillPluginsDefaults_Acme(t *testing.T) {
 			require.NoError(t, err)
 			require.NotNil(t, fullSchema)
 			assert.NoError(t, FillPluginsDefaults(plugin, fullSchema))
-			opts := cmpopts.IgnoreFields(*plugin, "Enabled", "Protocols", "storage_config", "allow_any_domain", "enable_ipv4_common_name", "preferred_chain", "rsa_key_size")
+			opts := cmpopts.IgnoreFields(*plugin, "Enabled", "Protocols")
 			if diff := cmp.Diff(plugin, tc.expected, opts); diff != "" {
 				t.Errorf(diff)
 			}
--- a/kong/utils_test.go
+++ b/kong/utils_test.go
@@ -1592,7 +1592,7 @@ func Test_FillPluginsDefaults_Acme(t *testing.T) {
 			require.NoError(t, err)
 			require.NotNil(t, fullSchema)
 			assert.NoError(t, FillPluginsDefaults(plugin, fullSchema))
-			opts := cmpopts.IgnoreFields(*plugin, "Enabled", "Protocols")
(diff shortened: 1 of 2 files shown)


aboudreault:
thanks @nishant95! I just tested your branch and I still have the same issue, my field is set to `null`.  My fix in #333 works understand that our PRs don't fix exactly the same thing. To support default with `set` of record, we'll need to merge both, right?

nishant95:
@aboudreault Yes, I raised this PR for issue #334 , this deals with record type only.
Your PR looks good and should fix it for type "set". 

But shouldn't there be a similar handling for type "map" as well? Similar to [this](https://github.com/Kong/kong/blob/750ba978b5ba9234d2ae6d84ac9806904c6cd9bc/kong/db/schema/init.lua#L1576) in Lua.

aboudreault:
@nishant95 this is possible yeah. Let's try to release this fix and #333, I will create a ticket for the map type.

rainest (approved):

rainest:
Overriding linter failure and merging cause it ain't worth waiting to round trip to the upstream branch and the error is inscrutable:

```
Check failure on line 2262 in kong/utils_test.go
GitHub Actions / lint

File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/kong) (gci)
```

golangci, please just print the actual command you need to run. Furthermore, the import block didn't change between those commits; it shouldn't start failing. This is not an import >:(

```
10:47:38-0700 esenin $ gci diff --skip-generated  -s standard -s default -s "prefix(github.com/kong)" utils_test.go 
--- utils_test.go
+++ utils_test.go
@@ -2259,4 +2259,4 @@
 			}
 		})
 	}
-}
\ No newline at end of file
+}
```

Going to fix in the release changelog commit.

kong/utils_test.go
```diff
@@ -1671,3 +2079,184 @@ func Test_FillPluginsDefaults_SetType(t *testing.T) {
 		})
 	}
 }
+
+func Test_FillPluginsDefaults_Acme(t *testing.T) {
+	RunWhenKong(t, "==3.3.0")
```
pmalek, line 2084: Is there a reason we want to run this against 3.3 **only**? If so then I believe this line deserves a comment why that's the case. It's not obvious to the reader why this particular version was chosen here and deserves a special treatment.

nishant95: @pmalek You are right, this check is not needed. This is a residual check which I added initially as the schema for this  plugin changed a lot over different kong versions, this is of no use now as the schema has been added to the test itself.

pmalek: Great! This should fix it then right? #342

nishant95: No, this should fix it: #343 Else the test would fail for many of the kong versions as the schema for Acme plugin would be different.

PR #338, opened 2023-06-06
avoid panic on bad Kong config

javierguerragiraldez:
if the Kong client fails to connect correctly, the RunWhenEnterprise function panics when reading the (missing) configuration.

Since this is a common failure in CI tests, this PR adds sanity tests to avoid the panic and get better failure logs.

### DIFF
--- a/kong/test_utils.go
+++ b/kong/test_utils.go
@@ -40,6 +40,22 @@ func RunWhenKong(t *testing.T, versionRange string) {
 	}
 }
 
+// shouldGet returns d[k], cast to type V.
+// if absent or the wrong type, fails the test instead of panic.
+func shouldGet[V any](t *testing.T, d map[string]any, k string) V {
+	v, ok := d[k]
+	if !ok {
+		t.Errorf("item '%s' not found", k)
+	}
+
+	v2, ok := v.(V)
+	if !ok {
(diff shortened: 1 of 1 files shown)


javierguerragiraldez:
depends on #337 to fix (remove) the `depguard` linter.  until that, no PR is likely to pass.

rainest (changes requested):
Should we not just

```
rawConfiguration, ok := info["configuration"]
if !ok {
    t.Fatal("could not retrieve Kong configuration")
}
configuration := rawConfiguration.(map[string]interface{})
```

as a simpler approach? It seems like the first will imply the latter two are also going to fail.

That aside, what's happening here where this isn't failing earlier? Shouldn't we get some failure [after the client request](https://github.com/Kong/go-kong/blob/86f46a5a816e10bda9dde660457f49566c4e9e7f/kong/test_utils.go#L72C25-L75)? Are we getting some issue where the root responds successfully, but with an empty response?

javierguerragiraldez:
> as a simpler approach? It seems like the first will imply the latter two are also going to fail.

yes, that should work too


> That aside, what's happening here where this isn't failing earlier?

not sure how it happens, but it does.  my guess is that sometimes kong isn't correctly loaded by the GiHub actions and the connection handshake doesn't complete.  the logs definitely point to a panic that happens here when there's no configuration.

PR #340, opened 2023-06-06
chore: add team-deck to CODEOWNERS

rainest:
In the course of https://github.com/Kong/go-kong/pull/339 I found that while there are team-deck members around during Americas hours, they're not actually allowed to approve go-kong PRs. They probably should be since the library is shared between teams.

Needs a main merge after https://github.com/Kong/go-kong/pull/339 to silence the linter.


pmalek (approved):
Needed to add write access for `Kong/team-deck` and only Github stopped complaining.

PR #342, opened 2023-06-07
tests: remove unneeded version constraint in `Test_FillPluginsDefaults_Acme`

pmalek:


nishant95:
Please use #343 instead.

PR #345, opened 2023-06-14
fix: handle empty array as nil when filling record defaults

aboudreault:
To avoid panic errors, in case of a wrong `[]` being set in the config for a record field, let's treat it as a nil and create the proper object map.

Can be reproduced with a deck sync:

```
_format_version: "3.0"
plugins:
- config:
    storage_config:
      redis: []
  enabled: true
  name: acme

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -253,11 +253,19 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 		frequired := value.Get(fname + ".required")
 		// Recursively fill defaults only if the field is either required or a subconfig is provided
 		if ftype.String() == "record" && (config[fname] != nil || (frequired.Exists() && frequired.Bool())) {
+			var fieldConfig Configuration
 			subConfig := config[fname]
-			if subConfig == nil {
-				subConfig = make(map[string]interface{})
+			switch subConfig.(type) {
+			case nil, []interface{}:
+				// We can encounter an empty array here due to an incorrect yaml
+				// setting or an empty subconfig (like acme.storage_config.kong).
+				// This should be treated as nil case.
+				// TODO: https://konghq.atlassian.net/browse/KOKO-1125
+				fieldConfig = make(map[string]interface{})
+			default:
+				fieldConfig = subConfig.(map[string]interface{})
 			}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: kong/utils.go [hunk 1]
```diff
@@ -253,11 +253,18 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
 		if ftype.String() == "record" && (config[fname] != nil || (frequired.Exists() && frequired.Bool())) {
+			var fieldConfig Configuration
 			subConfig := config[fname]
-			if subConfig == nil {
-				subConfig = make(map[string]interface{})
+			switch subConfig.(type) {
+			case nil, []interface{}:
+				// an empty array can be due to a wrong yaml setting or an
+				// empty subconfig (like acme.storage_config.kong) and should
+				// be treated as nil
```
pmalek, line 262:
```suggestion
				// We can encounter an empty array here due to an incorrect yaml
				// setting or an empty subconfig (like acme.storage_config.kong).
				// This should be treated as nil case.
				// TODO: <ref koko case>
```

pmalek (commented):
3 things I'd like to raise here:

- Add a reference of koko case in the code comment so that we know which issue this is related to
- This definitely deserves a changelog entry
- A test case should be added. There already exists [`Test_fillConfigRecord`](https://github.com/Kong/go-kong/blob/57f09c28c320257d8d1c4a1b52213f6ee6e9e120/kong/utils_test.go#L1659-L1728) so it should be as easy as adding another table driver test case there

kong/utils_test.go
```diff
@@ -1656,16 +1656,7 @@ func TestFillConsumerGroupPluginDefaults(T *testing.T) {
...
-	tests := []struct {
-		name     string
-		schema   gjson.Result
-		config   Configuration
-		expected Configuration
-	}{
-		{
-			name: "fills defaults for all missing fields",
-			schema: gjson.Parse(`{
+var fillConfigRecordTestSchema = `{
```
pmalek: 👍 just a small nit, this could be `const`

```suggestion
const fillConfigRecordTestSchema = `{
```

pmalek (approved):
Small nits, I applied those right away to speed up the delivery of this patch. 

Thanks for adding changelog and tests! 🙇

PR #348, opened 2023-06-15
fix: do not leak the response body on `Client.Do()`

pmalek:


rainest (approved):
This is breaking, correct? We keep the signature intact, but the exported type changes.

Pulling this into deck and KIC doesn't show any issues on vet though, so apparently we weren't using the response?

PR #352, opened 2023-06-23
feat: support scoping plugins to consumer-groups

GGabriele:


kong/plugin_service_test.go
```diff
@@ -739,6 +739,122 @@ func TestFillPluginDefaultsArbitraryMap(T *testing.T) {
...
+func TestPluginsWithConsumerGroup(T *testing.T) {
+	RunWhenDBMode(T, "postgres")
+	RunWhenEnterprise(T, ">=3.4.0", RequiredFeatures{})
+
+	assert := assert.New(T)
+	require := require.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 751:
```suggestion
	require.NoError(err)
	require.NotNil(client)
```

kong/plugin_service_test.go
```diff
@@ -739,6 +739,122 @@ func TestFillPluginDefaultsArbitraryMap(T *testing.T) {
...
+		plugin, err := client.Plugins.Create(defaultCtx, plugins[i])
+		assert.NoError(err)
+		assert.NotNil(plugin)
+		plugins[i] = plugin
+	}
+
+	pluginsFromKong, err := client.Plugins.ListAllForConsumerGroups(defaultCtx, createdCG.ID)
+	assert.NoError(err)
+	assert.NotNil(pluginsFromKong)
+	assert.Equal(2, len(pluginsFromKong))
```
pmalek, line 847:
```suggestion
	assert.Len(pluginsFromKong, 2)
```

kong/plugin_service_test.go
GGabriele: fixed

kong/plugin_service_test.go
GGabriele: fixed

kong/plugin_service.go
```diff
@@ -394,6 +419,16 @@ func (s *PluginService) ListAllForRoute(ctx context.Context,
 	return s.listAllByPath(ctx, "/routes/"+*routeID+"/plugins")
 }
 
+// ListAllForConsumerGroups fetches all Plugins in Kong enabled for a consumer grou.
```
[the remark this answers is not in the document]
GGabriele: fixed

kong/plugin_service.go
```diff
@@ -394,6 +419,16 @@ func (s *PluginService) ListAllForRoute(ctx context.Context,
 	return s.listAllByPath(ctx, "/routes/"+*routeID+"/plugins")
 }
 
+// ListAllForConsumerGroups fetches all Plugins in Kong enabled for a consumer group.
+func (s *PluginService) ListAllForConsumerGroups(ctx context.Context,
+	cgID *string,
```
pmalek, line 424: I wonder why we want to pass the param as pointer if the `nil` value is equivalent to an empty string? If it was indicating that it's optional then somewhat this maybe would have made sense but otherwise I'm puzzled why we do this.

kong/plugin_service.go
```diff
@@ -153,6 +157,27 @@ func (s *PluginService) CreateForRoute(ctx context.Context,
...
+// If an ID is specified, it will be used to
+// create a plugin in Kong, otherwise an ID
+// is auto-generated.
+func (s *PluginService) CreateForConsumerGroup(ctx context.Context,
+	cgIDorName *string, plugin *Plugin,
+) (*Plugin, error) {
+	queryPath := "/plugins"
+	method := "POST"
+
+	if plugin.ID != nil {
```
pmalek, line 170: Can `plugin` be nil? If that's the case this will blow up 💥

kong/plugin_service.go
```diff
@@ -153,6 +157,27 @@ func (s *PluginService) CreateForRoute(ctx context.Context,
...
+		queryPath = queryPath + "/" + *plugin.ID
+		method = "PUT"
+	}
+	if isEmptyString(cgIDorName) {
```
pmalek, line 174: Ditto about `cgIDorName` as pointer.

pmalek (commented):
Overall 👍 but I have some questions about the API decisions

kong/plugin_service.go
GGabriele: You're right. Fixed

pmalek (approved):

PR #354, opened 2023-06-30
feat: config helper

### LINKED ISSUE #353: Endpoint to retrieve current config
### Problem Statement

We want to have a new handler to fetch the current dataplane configuration.

### Proposed solution

A new handler needs to be added to interact with the admin `/config` path to retrieve the current dataplane configuration.

### Acceptance Criteria

- [ ] there is a way of retrieving the current dataplane configuration

mlavacca:
### What this PR does

This PR introduces a new helper to retrieve Kong's current config. The return value is an interface, and needs to be correctly parsed by the caller.

Fixes #353 

### Notes for your reviewer

This endpoint is only available when Kong is configured to not use a database, therefore it is tested only with Kong dbless.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,11 @@
 
 ## Unreleased
 
+> Release date: TBD
+
+- Added `Client.Config` handler, that can be used to obtain Kong's config.
+  [#354](https://github.com/Kong/go-kong/pull/354)
+
 ## [v0.44.0]
 
 > Release date: 2023/06/22
--- a/kong/client.go
+++ b/kong/client.go
@@ -3,6 +3,7 @@ package kong
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -355,6 +356,30 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
 		return nil, err
 	}
 	return &s, nil
+}
+
+// Config gets the specified config from the configured Admin API endpoint
+// and should contain the JSON serialized body that adheres to the configuration
+// format specified at:
+// https://docs.konghq.com/gateway/latest/production/deployment-topologies/db-less-and-declarative-config/#declarative-configuration-format
+// It returns the response body and an error, if it encounters any.
+func (c *Client) Config(ctx context.Context) ([]byte, error) {
+	req, err := c.NewRequest("GET", "/config", nil, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var configWrapper map[string]string
+	_, err = c.Do(ctx, req, &configWrapper)
+	if err != nil {
+		return nil, err
+	}
+	config, ok := configWrapper["config"]
+	if !ok {
+		return nil, errors.New("config field not found in GET /config response body")
+	}
+
+	return []byte(config), nil
 }
 
 // Root returns the response of GET request on root of Admin API (GET / or /kong with a workspace).
--- a/kong/client_test.go
+++ b/kong/client_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
 )
 
 func TestNewTestClient(t *testing.T) {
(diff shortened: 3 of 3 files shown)


kong/client.go
```diff
@@ -357,6 +357,22 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var config map[string]any
+	_, err = c.Do(ctx, req, &config)
+	if err != nil {
+		return nil, err
+	}
+	return (config)["config"], nil
```
czeslavo: I'm wondering whether it wouldn't be better to return raw config `[]byte`s instead of an interface. Having `[]byte` returned, we can tell the caller they have to unmarshal the underlying json (or yaml?). Returning just an interface, we return a nested go map which is already unmarshaled and transforming this into proper config structs will be painful on the caller side.

kong/client.go
```diff
@@ -357,6 +357,22 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
 	return &s, nil
 }
 
+// Config returns the config of a Kong node. The return value is an interface,
+// and needs to be parsed by the caller.
```
pmalek, line 361: Can we provide a schema struct that could be used to parse this return value to? I wonder if there isn't a known structure which is already available to be used in here for this purpose.

kong/client_test.go
```diff
@@ -32,6 +32,19 @@ func TestKongStatus(T *testing.T) {
 	assert.NotNil(status)
 }
 
+func TestKongConfig(t *testing.T) {
+	RunWhenDBMode(t, "off")
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
```
pmalek, line 40:
```suggestion
	require.NoError(t, err)
```

kong/client_test.go
```diff
@@ -32,6 +32,19 @@ func TestKongStatus(T *testing.T) {
...
+	assert.NotNil(client)
+
+	config, err := client.Config(defaultCtx)
+	assert.NoError(err)
+	assert.NotNil(config)
```
czeslavo, line 45: I think this would make sense to assert some structure of the returned config - e.g. if we go with returning `[]byte`, ensure it's a valid json in the test.

kong/client_test.go
```diff
@@ -32,6 +32,19 @@ func TestKongStatus(T *testing.T) {
...
 
+func TestKongConfig(t *testing.T) {
+	RunWhenDBMode(t, "off")
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
+
+	config, err := client.Config(defaultCtx)
```
pmalek, line 43: require.NoError(t, err)

kong/client_test.go
pmalek, line 45: Can't we make some assertions about this config struct? This could

- serve a documentation purpose to show readers how to use it
- guard us against unforeseen failures which `NotNil()` wouldn't

pmalek (commented):
I believe this would benefit from a Changelog entry.

Additionally maybe we could consider rephrasing the commit message to

```
feat: add Client.Config() which can obtain Kong's config
```

or similar. `config helper` seems a bit vague. WDYT?

kong/client.go
czeslavo: In KIC [`deck/types.Content`](https://github.com/Kong/deck/blob/43ac0a8426b315ab732f81c6423369fb924757be/file/types.go#L705) is used:  https://github.com/Kong/kubernetes-ingress-controller/blob/8d27fd1fa351be9a5e80204a37217a93bea9a5b4/internal/dataplane/sendconfig/inmemory.go#L58

czeslavo: I'd suggest making this comment include information similar to what we have for `ReloadDeclarativeRawConfig`: 

```go
// ... should contain the JSON
// serialized body that adheres to the configuration format specified at:
// https://docs.konghq.com/gateway/latest/production/deployment-topologies/db-less-and-declarative-config/#declarative-configuration-format
```

It could also mention the `deck/types.Content` as an example of such a struct for reference.

kong/client_test.go
mlavacca: I used "assert" to be coherent with all the other tests of this file.

CHANGELOG.md
```diff
@@ -1,63 +1,119 @@
 # Table of Contents
```
mlavacca, line 2: markdown formatter removed this space

kong/client_test.go
mlavacca: Assertions added

kong/client.go
mlavacca: The struct is available in Deck, as @czeslavo pointed out. I updated the comment as you suggested

kong/client.go
mlavacca: Yep, I agree. I returned `[]byte` so that it can be directly unmarshaled by the caller.

kong/client_test.go
pmalek: I get your point. OTOH though why copy broken patterns :)

pmalek (approved):

czeslavo (approved):

kong/client_test.go
mlavacca: Why broken? isn't just the same as using the approach you suggested?

pmalek: When `err` is not `nil` then most likely the `client` is `nil` (I'd suggest to change the check below to similarly use `require` as well).

In such a case any call to `client`'s methods would most likely panic. With `require` you wouldn't get a panic but a precise error message and the test would stop at this point.

mlavacca: Gotcha, I'm gonna fix it

mlavacca: fixed

PR #368, opened 2023-08-25
feat: add method Validate for RouteService

programmer04:
**What this PR does / why we need it:**

Gateway offers API for validating routes - `/schemas/routes/validate`. This PR extends `RouteService` with the method `Validate` (that calls aforementioned endpoint) and prepares the release of the new version `v0.47.0`.

**Which issue this PR fixes:**

Prerequisite to implement https://github.com/Kong/kubernetes-ingress-controller/issues/4557.

**Special notes for your reviewer:**

Hence this method will be used to validate expressions routes, extend tests with this configuration, and introduce a helper to skip tests when it is not applicable (also a small refactor to avoid unnecessary duplication). Extend the tests matrix with such a configuration, tests run quickly, but on the other hand, need a lot of runners - using `exclude` in workflow helps a lot. The number of checks increased from 45 to 66 (added Kong 3.4 to the matrix too).

Change in `kong/vault_service_test.go` is needed, because more values are in this map (for Kong 3.4) and others are rather random/irrelevant nad the region is presented for older versions, thus it doesn't break anything. 

**To consider:** creating an issue for covering with tests routes when expressions router is enabled.

### DIFF
--- a/.github/workflows/integration-test-enterprise.yaml
+++ b/.github/workflows/integration-test-enterprise.yaml
@@ -37,6 +37,9 @@ jobs:
     if: needs.secret-available.outputs.ok
     strategy:
       matrix:
+        router_flavor:
+          - 'traditional_compatible'
+          - 'expressions'
         kong_version:
         - '2.2'
         - '2.3'
@@ -50,6 +53,7 @@ jobs:
         - '3.2'
         - '3.3'
     env:
+      KONG_ROUTER_FLAVOR: ${{ matrix.router_flavor }}
       KONG_IMAGE_TAG: ${{ matrix.kong_version }}
       KONG_ANONYMOUS_REPORTS: "off"
       KONG_ADMIN_TOKEN: kong
--- a/.github/workflows/integration-test-enterprise.yaml
+++ b/.github/workflows/integration-test-enterprise.yaml
@@ -37,6 +37,24 @@ jobs:
     if: needs.secret-available.outputs.ok
     strategy:
       matrix:
+        # Skip explicitly to avoid spawning many unecessary jobs,
+        # since expressions router is supported for Kong >= 3.0.
+        # Option router_flavor is ignored for Kong < 3.0.
+        exclude:
+          - kong_version: '2.2'
+            router_flavor: 'expressions'
+          - kong_version: '2.3'
+            router_flavor: 'expressions'
+          - kong_version: '2.4'
+            router_flavor: 'expressions'
+          - kong_version: '2.5'
+            router_flavor: 'expressions'
+          - kong_version: '2.6'
+            router_flavor: 'expressions'
+          - kong_version: '2.7'
+            router_flavor: 'expressions'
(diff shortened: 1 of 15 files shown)


rainest (commented):
Is there a way to test this in isolation? It seems a bit silly to run all tests against both flavors when it only matters for the one in practice.

programmer04:
It's negligible that other test runs because the whole test suite (`make test`) takes ~10s. The problem (and bottleneck) is that router configuration can be only provided during the provisioning of the Kong Gateway and for every version and deployment type postgres, dbless, and enterprise new instance of Kong Gateway is spun up (instance of GitHub check). So even if I'd like to run only one test still the whole env has to be provisioned - an instance of GitHub check. Maybe it's worth improving it somehow, but I think it should create a separate issue for it

kong/vault_service_test.go
```diff
@@ -75,7 +75,9 @@ func TestVaultsService(t *testing.T) {
 	require.Equal(id, *createdVault.ID)
 	require.Equal("aws", *createdVault.Name)
 	require.Equal("aws vault for secrets", *createdVault.Description)
-	require.Equal(Configuration{"region": "us-east-2"}, createdVault.Config)
+	region, ok := createdVault.Config["region"]
+	require.Equal(true, ok)
```
czeslavo, line 79:
```suggestion
	require.True(ok)
```

.github/workflows/integration-test.yaml
```diff
@@ -19,23 +19,50 @@ jobs:
...
-        - '2.4'
-        - '2.5'
-        - '2.6'
-        - '2.7'
-        - '2.8'
-        - '3.0'
-        - '3.1'
-        - '3.2'
-        - '3.3'
+        # Skip explicitly to avoid spawning many unecessary jobs,
```
czeslavo, line 22:
```suggestion
        # Skip explicitly to avoid spawning many unnecessary jobs,
```

.github/workflows/integration-test-enterprise.yaml
```diff
@@ -37,6 +37,27 @@ jobs:
     if: needs.secret-available.outputs.ok
     strategy:
       matrix:
+        # Skip explicitly to avoid spawning many unecessary jobs,
```
czeslavo, line 40:
```suggestion
        # Skip explicitly to avoid spawning many unnecessary jobs,
```

kong/route_service_test.go
```diff
@@ -302,3 +310,51 @@ func TestRouteWithHeaders(T *testing.T) {
 	err = client.Routes.Delete(defaultCtx, createdRoute.ID)
 	assert.NoError(err)
 }
+
+func TestRoutesValidation(T *testing.T) {
```
czeslavo, line 314: I think we should also add a test case for `traditional_compatible` (e.g. invalid regex).

czeslavo (approved):

PR #374, opened 2023-10-17
fix: don't overwrite fields set to empty arrays when default exists

GGabriele:
This commit fixes a bug preventing users to set fields to emtpy arrays when a default for those fields exist. For example, imagine we have an `openid-connect` plugin configured via the UI having the `login_tokens` set to an empty array (note: this field has a default value of `["id_token"]`:

```
$ http :8001/plugins/$id | jq .config.login_tokens
[]
```

If we do a `deck dump`, the configuration will be reflected into a local state file:

```
$ deck dump --yes

$ cat kong.yaml | grep login_tokens
    login_tokens: []
```

But, if we now run `deck diff` we see that decK will detect a difference because of the `login_tokens` field's default value:

```
$ deck diff
updating plugin openid-connect (global)  {
   "config": {
     "anonymous": null,
     ...
     ...
     "login_redirect_mode": "fragment",
     "login_redirect_uri": null,
     "login_tokens": [
+      "id_token"
     ],
     "logout_methods": [
       "POST",
       "DELETE"
     ],
     ...
     ...
 }

Summary:
  Created: 0
  Updated: 1
  Deleted: 0
```

This commit corrects this defect by allowing decK to set empty arrays when default values exist for a given field.

### DIFF
--- a/kong/utils_test.go
+++ b/kong/utils_test.go
@@ -2330,12 +2330,12 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
 	}
 
 	for _, tc := range tests {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			plugin := tc.plugin
 			var fullSchema map[string]interface{}
-			err := json.Unmarshal([]byte(NonEmptyDefaultArrayFieldSchema), &fullSchema)
+			require.NoError(t, json.Unmarshal([]byte(NonEmptyDefaultArrayFieldSchema), &fullSchema))
 
-			require.NoError(t, err)
 			require.NotNil(t, fullSchema)
 			assert.NoError(t, FillPluginsDefaults(plugin, fullSchema))
 			opts := cmpopts.IgnoreFields(*plugin, "Enabled", "Protocols")
--- a/kong/utils_test.go
+++ b/kong/utils_test.go
@@ -2212,3 +2212,136 @@ func Test_FillPluginsDefaults_Acme(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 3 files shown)


kong/utils_test.go
```diff
@@ -2212,3 +2212,136 @@ func Test_FillPluginsDefaults_Acme(t *testing.T) {
...
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			plugin := tc.plugin
+			var fullSchema map[string]interface{}
+			err := json.Unmarshal([]byte(NonEmptyDefaultArrayFieldSchema), &fullSchema)
+
+			require.NoError(t, err)
```
pmalek, line 2338:
```suggestion
			require.NoError(t, json.Unmarshal([]byte(NonEmptyDefaultArrayFieldSchema), &fullSchema))
```

kong/utils_test.go
```diff
@@ -2212,3 +2212,136 @@ func Test_FillPluginsDefaults_Acme(t *testing.T) {
...
+					"issuer":       "https://accounts.google.com",
+					"login_tokens": []any{"access_token", "refresh_token"},
+				},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			plugin := tc.plugin
```
pmalek, line 2334:
```suggestion
	for _, tc := range tests {
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
			plugin := tc.plugin
```

pmalek (commented):
Some nits, but overall 👍 

I believe we want to add a changelog entry for this, right?

pmalek (approved):

PR #378, opened 2023-10-27
fix: update Route's priority type to uint64

pmalek:
Route's priority in expression routes should be of type `uint64` instead of int.

---

There's a related (linked in the comment) issue for Gateway which returns integers in scientific notation: https://konghq.atlassian.net/browse/FTI-5515. When that's solved the commented test case should be uncommented.

---

Also update the changelog to release 0.48.0.

### DIFF
--- a/kong/route.go
+++ b/kong/route.go
@@ -14,7 +14,7 @@ type Route struct {
 	Paths         []*string           `json:"paths,omitempty" yaml:"paths,omitempty"`
 	PathHandling  *string             `json:"path_handling,omitempty" yaml:"path_handling,omitempty"`
 	PreserveHost  *bool               `json:"preserve_host,omitempty" yaml:"preserve_host,omitempty"`
-	Priority      *int                `json:"priority,omitempty" yaml:"priority,omitempty"`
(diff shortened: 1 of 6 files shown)


kong/route.go
```diff
@@ -14,7 +14,7 @@ type Route struct {
 	Paths         []*string           `json:"paths,omitempty" yaml:"paths,omitempty"`
 	PathHandling  *string             `json:"path_handling,omitempty" yaml:"path_handling,omitempty"`
 	PreserveHost  *bool               `json:"preserve_host,omitempty" yaml:"preserve_host,omitempty"`
-	Priority      *int                `json:"priority,omitempty" yaml:"priority,omitempty"`
+	Priority      *int64              `json:"priority,omitempty" yaml:"priority,omitempty"`
```
czeslavo, line 17: Isn't it technically a breaking change? It will break if someone uses it like:

```go
kong.Route{ 
    Priority: kong.Int(1),
}
```

I think it should be a minor bump at least + a release note entry, WDYT?

czeslavo (approved):

randmonkey (approved):

PR #391, opened 2023-12-28
feat: add FillID() method for vaults

### LINKED ISSUE #390: Add `FillID` method for vaults
Currently vaults does not support `FillID()` method, so KIC cannot generate an ID when creating a Kong vault. Now we need to implement a `FillID()` method for `Vault`.

randmonkey:
Add `FillID` method for `Vault`. Fixes #390.

### DIFF
--- a/kong/ids.go
+++ b/kong/ids.go
@@ -104,7 +104,8 @@ func (cg *ConsumerGroup) FillID() error {
 }
 
 // FillID fills the ID of an entity. It is a no-op if the entity already has an ID.
-// ID is generated in a deterministic way using UUIDv5. The UUIDv5 namespace is different for each entity type.
+// ID is generated in a deterministic way using UUIDv5.
+// The UUIDv5 namespace being used for generation is separate from other namespaces used for generating IDs for other types.
 // The name used to generate the ID for Vault is Vault.Prefix.
 func (v *Vault) FillID() error {
 	if v == nil {
--- a/kong/ids.go
+++ b/kong/ids.go
@@ -110,11 +110,11 @@ func (v *Vault) FillID() error {
 	if v == nil {
 		return fmt.Errorf("vault is nil")
 	}
-	if v.ID != nil && *v.ID != "" {
+	if v.ID != nil && len(*v.ID) > 0 {
 		// ID already set, do nothing.
 		return nil
 	}
-	if v.Prefix == nil || *v.Prefix == "" {
+	if v.Prefix == nil || len(*v.Prefix) == 0 {
 		return fmt.Errorf("vault prefix is required")
 	}
 
--- a/kong/ids.go
+++ b/kong/ids.go
@@ -103,6 +103,30 @@ func (cg *ConsumerGroup) FillID() error {
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


kong/ids.go
```diff
@@ -103,6 +103,30 @@ func (cg *ConsumerGroup) FillID() error {
...
+// The name used to generate the ID for Vault is Vault.Prefix.
+func (v *Vault) FillID() error {
+	if v == nil {
+		return fmt.Errorf("vault is nil")
+	}
+	if v.ID != nil && *v.ID != "" {
+		// ID already set, do nothing.
+		return nil
+	}
+	if v.Prefix == nil || *v.Prefix == "" {
```
pmalek, line 117: We could probably check this with `len(v.Prefix) == 0` as well with the same effect.

kong/ids.go
```diff
@@ -103,6 +103,30 @@ func (cg *ConsumerGroup) FillID() error {
...
 }
 
+// FillID fills the ID of an entity. It is a no-op if the entity already has an ID.
+// ID is generated in a deterministic way using UUIDv5. The UUIDv5 namespace is different for each entity type.
+// The name used to generate the ID for Vault is Vault.Prefix.
+func (v *Vault) FillID() error {
+	if v == nil {
+		return fmt.Errorf("vault is nil")
+	}
+	if v.ID != nil && *v.ID != "" {
```
pmalek, line 113: ditto

kong/ids.go
```diff
@@ -103,6 +103,30 @@ func (cg *ConsumerGroup) FillID() error {
 	return nil
 }
 
+// FillID fills the ID of an entity. It is a no-op if the entity already has an ID.
+// ID is generated in a deterministic way using UUIDv5. The UUIDv5 namespace is different for each entity type.
```
pmalek, line 107: > The UUIDv5 namespace is different for each entity type.

What meaning does this have in here? I believe we can drop this sentence from the godoc.

randmonkey: This is just copied from `FillID` method of other entities. I think this sentence here is used to tell that different entities with the same (entity wide unique) key will have different IDs because we use UUIDv5 and assign different namespaces for each kind of entities.

kong/ids.go
randmonkey: Done.

kong/ids.go
randmonkey: Done.

kong/ids.go
pmalek: Got it. But in the context of this method (which we're discussing) this sentence doesn't really make much sense does it?

This method is only relevant for `Vault`s, so other types (mentioned in the godoc here) do not make sense IMHO.

But as I look at it now, I believe the concealed shared bit of this method (shared map of generators) might have this effect 🤔 

Perhaps using the following wording would make more sense?

```suggestion
// ID is generated in a deterministic way using UUIDv5. The UUIDv5 namespace being used for generation is separate from other namespaces used for generating IDs for other types.
```

randmonkey: Done.

pmalek (approved):

PR #392, opened 2023-12-28
add validate method in vault service

randmonkey:
Add `Validate()` method to call `/schemas/vaults/validate` to validate vaults.

### DIFF
--- a/kong/vault_service.go
+++ b/kong/vault_service.go
@@ -3,7 +3,9 @@ package kong
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
 )
 
 // AbstractVaultService handles Vaults in Kong.
@@ -20,6 +22,8 @@ type AbstractVaultService interface {
 	List(ctx context.Context, opt *ListOpt) ([]*Vault, *ListOpt, error)
 	// ListAll fetches all Vaults in Kong.
 	ListAll(ctx context.Context) ([]*Vault, error)
+	// Validate validates a Vault against its schema.
+	Validate(ctx context.Context, vault *Vault) (bool, string, error)
 }
 
 // VaultService handles Vaults in Kong.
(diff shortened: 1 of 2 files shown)


kong/vault_service.go
```diff
@@ -152,3 +156,28 @@ func (s *VaultService) ListAll(ctx context.Context) ([]*Vault, error) {
 	}
 	return vaults, nil
 }
+
+func (s *VaultService) Validate(ctx context.Context, vault *Vault) (bool, string, error) {
+	endpoint := "/schemas/vaults/validate"
```
pmalek, line 161: Can we make this a `const`?

kong/vault_service.go
```diff
@@ -152,3 +156,28 @@ func (s *VaultService) ListAll(ctx context.Context) ([]*Vault, error) {
 	}
 	return vaults, nil
 }
+
+func (s *VaultService) Validate(ctx context.Context, vault *Vault) (bool, string, error) {
```
programmer04, line 160: Could it be written like a Validate [here](https://github.com/Kong/go-kong/blob/abc671e70baee66723ed806889930d07b65f2a67/kong/route_service.go#L216C1-L230)? I think it's more consistent

pmalek: 👍 Although I'd mention what the return values are. One could deduce what the `bool` is for but string isn't obvious here IMHO.

programmer04: Yup it ought to be explained in docstring

kong/vault_service.go
randmonkey: done

kong/vault_service.go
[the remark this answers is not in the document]
randmonkey: done.

randmonkey: Done.

programmer04 (approved):

kong/vault_service.go
```diff
@@ -152,3 +156,23 @@ func (s *VaultService) ListAll(ctx context.Context) ([]*Vault, error) {
 	}
 	return vaults, nil
 }
+
+// Validate validates a vault against its schema.
+// returns validate result (passed/failed) and the message from the schema validation service if validation fails.
+// returns a non-nil error if failed to call the schema validation service.
```
pmalek, line 162: `returns` -> `Returns`

Issue #393, opened 2024-01-03
Consider workspace in FillID method


randmonkey:
As investigated in https://github.com/Kong/kubernetes-ingress-controller/issues/5389, the FillID() method may generate the same ID for different entities with same "key" in different workspaces. However, ID should be unique through ALL workspaces. This will bring conflicts in configurations.

 For example, if two consumers in different workspaces have  the same `username`, we will generate the same ID for the two consumers. Then one of them will receive a "ID already exist" error as described in https://github.com/Kong/kubernetes-ingress-controller/issues/5389. To resolve this, we need to consider workspace of the entity to generate ID from.

randmonkey:
fixed in #395 .

PR #395, opened 2024-01-05
add workspace param in FillID methods

randmonkey:
Add `workspace` parameter in `FillID` methods of supporting entities to avoid ID conflict on entities in different namespaces with the same "key". fixes #394.


programmer04 (approved):
Probably the wrong issue referenced in the description of this PR, should be #393

PR #399, opened 2024-01-12
extract details from response

### LINKED ISSUE #398: Parse `details` field in error response from Konnect of entity CRUD requests
required by https://github.com/Kong/kubernetes-ingress-controller/issues/5274. 
When an entity CRUD request to Konnect fails, the response would include a `details` field to tell the details of the failure, for exmple
```
{"code":3,"message":"validation error","details":[{"@type":"type.googleapis.com/kong.admin.model.v1.ErrorDetail","type":"ERROR_TYPE_FIELD","field":"value.rsa_public_key","messages":["'-----BEGIN RSA PUBLIC KEY-----\\nMIIREDACTED...\\n-----END RSA PUBLIC KEY-----' is not valid 'pem-pkix-encoded-public-key'"]}]}
```
Now the type `APIError` includes a `details` field, we could add the response content of `details` field from Konnect.

randmonkey:
Extract `details` field from response if error happens in calling admin API on Konnect to do CRUD of entities. fixes #398.

### DIFF
--- a/kong/response.go
+++ b/kong/response.go
@@ -37,6 +37,18 @@ func messageFromBody(b []byte) string {
 	return s.Message
 }
 
+// detailsFromBodyDetailsField extract details from body if the response body contains a "details" field.
+// Used for extracting details from response from Konnect APIs when error happens.
+func detailsFromBodyDetailsField(b []byte) any {
+	s := struct {
+		Details any `json:"details"`
+	}{}
+	if err := json.Unmarshal(b, &s); err != nil {
+		return nil
+	}
+	return s.Details
+}
+
 func hasError(res *http.Response) error {
 	if res.StatusCode >= 200 && res.StatusCode <= 399 {
 		return nil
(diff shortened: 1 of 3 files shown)


kong/response.go
```diff
@@ -48,14 +60,18 @@ func hasError(res *http.Response) error {
...
+	if details, ok := extractErrDetails(res, body); ok {
 		apiErr.SetDetails(details)
 	}
 
 	return apiErr
 }
 
-func extractErrDetails(res *http.Response) (any, bool) {
+func extractErrDetails(res *http.Response, body []byte) (any, bool) {
+	// Firstly extract `details` field from response.
```
czeslavo, line 71: I think this could potentially break the expected handling of `429` from Konnect. This could happen in a situation where we get a response with `429` code and `details` filled. 

Because of that, I think that we should rather change the order here and first handle the specific cases based on the status code, and only fallback to generic details once we have no logic for a given case.

randmonkey: Put extracting from body after status code process.

czeslavo (approved):

PR #400, opened 2024-01-17
add details in APIError.Error

randmonkey:
Because `sync.Solve` wraps the error returned from CRUD actions by `fmt.Errorf()`, the `details` which are not included in `APIError.Error()` method is lost in wrapping errors. So https://github.com/Kong/kubernetes-ingress-controller/issues/5274 requires `details` to be included in `Error()` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,8 +65,10 @@
 ## Unreleased
 
 - Added extract of `details` field in response when CRUD API fails to extract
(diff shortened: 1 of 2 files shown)


czeslavo (approved):

randmonkey:
Close this because we can extract `APIError` from errors returned in `sync.Solve` in KIC.

PR #405, opened 2024-01-25
feat: add ConfigSource field into ConsumerGroupPlugin

GGabriele:
This commit is adding a new `ConfigSource` field into the `ConsumerGroupPlugin` struct. This is not a field that should be used for online operation, but it's needed by the the `go-database-reconciler` to allow decK to work with plugins config deduplication with Consumer Groups as well.

Ref: https://docs.konghq.com/deck/latest/guides/deduplicate-plugin-configuration/


rainest (approved):
I am somewhat amused by the notion of deduplicating configuration on a resource that exists expressly for deduplicating configuration on a different type of resource.

PR #421, opened 2024-02-26
chore(*) fix lints

rainest:
Blocking dependabot upgrades after a linter upgrade, e.g. https://github.com/Kong/go-kong/actions/runs/8045072066/job/21980638584?pr=420

Discard a bunch of unused parameters and rename a poorly-capitalized argument.

Could probably discard most of the args entirely, but :shrug: they're mostly test objects in test functions, which are kinda standard-ish, so I kept the placeholder around.

### DIFF
--- a/kong/route_service_test.go
+++ b/kong/route_service_test.go
@@ -423,7 +423,7 @@ func TestRoutesValidationExpressions(T *testing.T) {
 			valid: true,
 		},
 	} {
-		T.Run(tC.name, func(t *testing.T) {
+		T.Run(tC.name, func(_ *testing.T) {
 			ok, msg, err := client.Routes.Validate(defaultCtx, tC.route)
 			require.NoError(err)
 			require.Equal(tC.valid, ok)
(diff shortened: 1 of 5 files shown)


flrgh (approved):

kong/route_service_test.go
```diff
@@ -492,7 +492,7 @@ func TestRoutesValidationTraditionalCompatible(T *testing.T) {
 			msgStartWith: "schema violation (paths.2: invalid regex:",
 		},
 	} {
-		T.Run(tC.name, func(t *testing.T) {
```
pmalek, line 495: We should use that `t` in assertions below to fail the subtest and not the whole suite

pmalek (changes requested):

rainest:
AFAICT https://github.com/Kong/go-kong/commit/241dbb8ae59e27b2a803a02ca19b23ffe64ccb5b includes all of what was originally here.

PR #430, opened 2024-03-27
feat: support setting custom headers

szesch:
Currently to set request headers `HTTPClientWithHeaders` is used to configure an `http.Client` that injects headers on each request via a custom `Transport`. This is problematic when there is a desire to reuse the underlying `http.Client`. This PR adds support for capturing the desired headers when creating the Kong `Client` and then `Do` sets the headers.

If users are OK with a customized `http.Client` then they can continue using `NewClient` with `HTTPClientWithHeaders`. If they wish to share the underlying `http.Client` then they can use the new function `NewClientWithHeaders`.


rainest (commented):
Can you clarify the problem scenario with what you're trying to do and `HTTPClientWithHeaders()`? What is the client being shared with and why?

Ideally we'd avoid a proliferation of similar public APIs, though I expect we've probably written ourselves into a bit of corner with the existing code and can't easily have a single option without a breaking refactor.

---

For background, we have a bit of a mess here where deck/GDR and the controller both add headers, but do so differently.

deck [uses](https://github.com/Kong/go-database-reconciler/blob/225a74d58a05d5e1fb0677ba0183e03381088948/pkg/utils/types.go#L235) the existing [go-kong client with headers utility](https://github.com/Kong/go-kong/blob/241dbb8ae59e27b2a803a02ca19b23ffe64ccb5b/kong/utils.go#L78-L133).

KIC uses its own internal [client with headers](https://github.com/Kong/kubernetes-ingress-controller/blob/e237ebc22b606e9680c5230131dad827c8706ecd/internal/adminapi/kong.go#L184-L188) implementation, which it then [passes to go-kong](https://github.com/Kong/kubernetes-ingress-controller/blob/e237ebc22b606e9680c5230131dad827c8706ecd/internal/adminapi/kong.go#L47). This implementation looks roughly the same as the go-kong utility (it builds a RoundTripper implementation with a header list and then stuffs those into the request when clients call `RoundTrip()`).

I don't know if there are any different requirements between the two that necessitate the split (I doubt it) or if there's any reason they'd have issues using the new implementation (adding headers per-request in `kong.Client.Do()`) proposed here. The setup signatures for the existing and proposed APIs both accept an `http.Client` and `http.Headers`. IDK what'd all be required in the refactor to handle the `baseURL` and different return types (the existing util returns an `http.Client`, the proposed one returns a `kong.Client`).

szesch:
I would like to do something like this:
```Go
httpClient := &http.Client{ 
  // some configuration
}

for _, o := range konnectOrgs {
  headers := codeToGetAuthHeader(o)
  kongClient := kong.HTTPClientWithHeaders(baseURL, httpClient, headers)
 
  // Now manage some entities with kongClient
}
```

That is a simplified example, but I basically want a long lived http client object that I can reuse across kong clients as well as other HTTP based clients. I can't do this today since `HTTPClientWithHeaders` modifies the `Transport`. Another way to think of it is that I want custom headers tied to the kong client and not the http client.

rainest:
Gotcha, what's the motivation for reusing the client for non-Kong requests? What's lost if you need to build the other clients separately

Sharing the `http.Client` for other purposes isn't something we really intended here--the go-kong client should be an opinionated collection of settings that are common across all requests to a particular Kong cluster.

We do expose the `http.Client` as an input but it's not clear why we do so instead of always building the client on behalf of the downstream application. My best guess is that we just didn't do a good job of refactoring delineation of responsibilities when we broke go-kong out into a library. If anything we'd want to make it more opinionated about the `Transport` configuration going forward.

kong/client.go
```diff
@@ -107,6 +112,12 @@ type Status struct {
 	ConfigurationHash string `json:"configuration_hash,omitempty" yaml:"configuration_hash,omitempty"`
 }
 
+func WithDoer(doer Doer) Option {
```
rainest, line 115:
```suggestion
// WithDoer sets the Client's request dispatcher, the function that sends a request to Kong and returns its response.
func WithDoer(doer Doer) Option {
```

In addition to the comment, any reason not to make this a method? That's more typical for the various setters, such as `Client.SetWorkspace()`.

Also, since I can't comment on the unmodified line, please update the comment for `Client.Do()` to mention the override:

```

// Do executes an HTTP request and returns a Response.
//
// The caller can optionally provide v parameter, which when provided will contain
// the response body. Do supports wither an io.Writer (which will contain the
// response body verbatim) or anything else which the body should be unmarshalled
// into.
//
// By default, Do() calls DoRaw() to send the request and return the response before unmarshalling, logging,
// and error handling. The Client's WithDoer() method allows overriding this to inject custom behavior.
```

rainest (changes requested):
I like the more general "insert custom advanced functionality here" approach better than the second header set method, though don't know that it'd need the Options handler to arbitrarily modify any client field. Do we have a known use case for that other than the custom `DoRaw()` at present? If not I'd recommend sticking with a dedicated function that sets the `Client.doer` only.

rainest (approved):

PR #436, opened 2024-04-08
chore: validate type assertion to prevent panic

reversearrow:
I observed panics couple of times in tests. Validating type assertion before proceeding further should prevent this panic.


rainest (approved):
This is presumably caused by the test checking the admin API before it's properly available. We should try to make that not happen, but I suppose this is at least slightly better than the panic.

PR #443, opened 2024-05-10
feat: add SchemaService.Validate to validate arbitrary entity

randmonkey:
Add `SchemaService.Validate` to call `POST /schemas/{entity type}/validate` to validate arbitrary type of Kong entity. Required for validating custom entities against Kong gateway in KIC https://github.com/Kong/kubernetes-ingress-controller/issues/5975.

### DIFF
--- a/kong/schema_service.go
+++ b/kong/schema_service.go
@@ -7,13 +7,31 @@ import (
 	"net/http"
 )
 
+// EntityType describes a type of Kong entity.
+type EntityType string
+
+// Defition of constants for standard types of Kong entities.
+const (
+	EntityTypeServices       EntityType = "services"
+	EntityTypeRoutes         EntityType = "routes"
+	EntityTypeUpstreams      EntityType = "upstreams"
+	EntityTypeTargets        EntityType = "targets"
+	EntityTypePlugins        EntityType = "plugins"
+	EntityTypeConsumers      EntityType = "consumers"
+	EntityTypeConsumerGroups EntityType = "consumer_groups"
+	EntityTypeSNIs           EntityType = "snis"
+	EntityTypeCertificates   EntityType = "certificates"
+	EntityTypeCACertificates EntityType = "ca_certificates"
+	EntityTypeTags           EntityType = "tags"
+)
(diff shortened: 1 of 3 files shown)


kong/schema_service.go
```diff
@@ -2,13 +2,18 @@ package kong
...
 	"fmt"
+	"net/http"
 )
 
 // AbstractSchemaService handles schemas in Kong.
 type AbstractSchemaService interface {
 	// Get fetches an entity schema from Kong.
 	Get(ctx context.Context, entity string) (Schema, error)
+	// Validate validates an arbitrary entity in Kong.
+	// Used for custom entities, or entities that does not have a Validate method in the corresponding service.
```
czeslavo, line 15:
```suggestion
	// Used for custom entities, or entities that do not have a Validate method in the corresponding service.
```

kong/schema_service.go
```diff
@@ -2,13 +2,18 @@ package kong
...
+	Validate(ctx context.Context, entityType string, entity interface{}) (bool, string, error)
```
czeslavo, line 16: Just food for thought: `entityType` passed as an arbitrary string is quite error-prone - users can pass e.g. `service` instead of `services`. That is not obvious what are the allowed entity types here. We could define an enum that would include all the allowed entities and map them to the endpoint name (`services`, `vaults`, etc.). On the other hand, I understand this has to allow arbitrary names as it has to support custom entities with names we do not know upfront - from this perspective, a rigid enum doesn't make sense.

We could meet somewhere between and define const strings for the built-in types, still accepting an arbitrary `entityType string`. The question is whether this is something we need, taking into account that we already have strongly typed variants of `Validate` in particular built-in entities' service interfaces.

kong/schema_service.go
randmonkey: done.

kong/schema_service.go
randmonkey: Defined `EntityType` and added constants for some known types.

czeslavo (approved):

PR #444, opened 2024-05-10
test: update priority in route to `2^46 - 1`

randmonkey:
Since Kong gateway will limit the `priority` in routes to be less than `2^46` (in `[0,2^46-1]`) , the `priority`of route used in tests should be updated to be in the interval.

### DIFF
--- a/kong/route_service_test.go
+++ b/kong/route_service_test.go
@@ -418,7 +418,7 @@ func TestRoutesValidationExpressions(T *testing.T) {
 			name: "valid expression with priority",
 			route: &Route{
 				Expression: String(`lower(http.path) ^= "/prefix/"`),
-				Priority:   Uint64((1 << 46) - 1),
+				Priority:   Uint64(MaxRoutePriority),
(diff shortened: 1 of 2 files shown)


kong/route_service_test.go
```diff
@@ -418,7 +418,7 @@ func TestRoutesValidationExpressions(T *testing.T) {
 			name: "valid expression with priority",
 			route: &Route{
 				Expression: String(`lower(http.path) ^= "/prefix/"`),
-				Priority:   Uint64(3382097767104500),
+				Priority:   Uint64((1 << 46) - 1),
```
czeslavo, line 421: How about defining this as a public const (e.g. `MaximalRoutePriority` so that `go-kong` users can determine the maximal priority value? Then, we could also use this here, in the test, instead of hardcoding.

randmonkey: Moved to a const.

czeslavo (approved):

PR #454, opened 2024-07-17
feat: add /status/ready function

tao12345666333:
ref: https://github.com/Kong/kubernetes-ingress-controller/issues/5068

> This endpoint should be used to health-check Kong nodes. This can be health checks from orchestration frameworks like k8s or by load-balancers fronting Kong nodes that proxy traffic. This endpoint returns 200 only after the Kong node has configured itself and is ready to start proxying traffic.

https://docs.konghq.com/gateway/api/status/latest/#/default/get_status_ready

### DIFF
--- a/kong/client.go
+++ b/kong/client.go
@@ -11,6 +11,8 @@ import (
 	"net/http/httputil"
 	"net/url"
 	"os"
+	"regexp"
+	"strings"
 	"sync"
 	"time"
 
@@ -18,7 +20,14 @@ import (
 )
 
 const (
+	// ref: https://docs.konghq.com/gateway/latest/production/networking/default-ports/
+	// defaultBaseURL is the endpoint for admin API
 	defaultBaseURL = "http://localhost:8001"
+	// defaultStatusURL is the endpoint for status API
+	// By default, the Status API listens on 127.0.0.1
+	// If you need to request it from elsewhere,
+	// please modify the `KONG_STATUS_LISTEN` environment variable of Gateway.
+	defaultStatusURL = "http://localhost:8007"
 	// DefaultTimeout is the timeout used for network connections and requests
 	// including TCP, TLS and HTTP layers.
 	DefaultTimeout = 60 * time.Second
@@ -40,6 +49,7 @@ var defaultCtx = context.Background()
 type Client struct {
 	client                  *http.Client
 	baseRootURL             string
+	statusURL               string
 	workspace               string       // Do not access directly. Use Workspace()/SetWorkspace().
 	UserAgent               string       // User-Agent for the client.
 	workspaceLock           sync.RWMutex // Synchronizes access to workspace.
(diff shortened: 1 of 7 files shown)


kong/client.go
```diff
@@ -11,14 +11,23 @@ import (
...
+	"strings"
 	"sync"
 	"time"
 
 	"github.com/kong/go-kong/kong/custom"
 )
 
 const (
+	// ref: https://docs.konghq.com/gateway/latest/production/networking/default-ports/
+	// defaultBaseURL is the endpoint for admin API
```
pmalek, line 24:
```suggestion
	// defaultBaseURL is the endpoint for admin API.
	// ref: https://docs.konghq.com/gateway/latest/production/networking/default-ports/	
```

kong/client.go
```diff
@@ -11,14 +11,23 @@ import (
...
 	defaultBaseURL = "http://localhost:8001"
+	// defaultStatusURL is the endpoint for status API
+	// By default, the Status API listens on 127.0.0.1
```
pmalek, line 27:
```suggestion
	// By default, the Status API listens on localhost.
```

kong/client.go
```diff
@@ -111,8 +121,39 @@ type Status struct {
...
+	Message string `json:"message"`
+}
+
+type RequestOptions struct {
+	BaseURL   *string
+	StatusURL *string
+}
+
+func parseStatusListen(listen string) string {
+	re := regexp.MustCompile(`^([\w\.:]+)\s*(.*)?`)
```
pmalek, line 134: Can we extract the regex to not compile it on every call?

kong/client.go
```diff
@@ -111,8 +121,39 @@ type Status struct {
...
+	matches := re.FindStringSubmatch(listen)
+
+	if len(matches) == 0 {
+		return ""
+	}
+
+	address := matches[1]
+	extraParams := matches[2]
```
pmalek, line 142: What if `matches` is 1 or 2 entries long? This will blow up then. Can we check for that as well here?

kong/client.go
```diff
@@ -391,6 +455,21 @@ func (c *Client) Status(ctx context.Context) (*Status, error) {
 	return &s, nil
 }
 
+// Ready returns 200 only after the Kong node has configured itself and is ready to start proxying traffic.
```
pmalek, line 458: Can we link to the official docs here? Specifically to https://docs.konghq.com/gateway/latest/production/monitoring/healthcheck-probes/#types-of-health-checks

I'd consider mentioning that this function uses Kong's "Readiness" endpoint.

kong/client_test.go
```diff
@@ -34,6 +35,53 @@ func TestKongStatus(T *testing.T) {
...
+	}
+
+	assert := assert.New(t)
+
+	client, err := NewTestClientWithOpts(RequestOptions{
+		BaseURL:   String("http://localhost:8001"),
+		StatusURL: String("http://localhost:8100"),
+	}, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 72:
```suggestion
	assert := assert.New(t)
	require := require.New(t)

	client, err := NewTestClientWithOpts(
		RequestOptions{
			BaseURL:   String("http://localhost:8001"),
			StatusURL: String("http://localhost:8100"),
		}, nil,
	)
	require.NoError(err)
	require.NotNil(client)
```

kong/client_test.go
```diff
@@ -34,6 +35,53 @@ func TestKongStatus(T *testing.T) {
...
+
+	sm, err := client.Ready(defaultCtx)
+	if err != nil {
+		// for dbless mode, the ready endpoint returns 503
```
pmalek, line 76: WDYT about moving this message to assertion so that it's logged when the test fails?

kong/client_test.go
```diff
@@ -34,6 +35,53 @@ func TestKongStatus(T *testing.T) {
...
+		assert.Equal("HTTP status 503 (message: \"no configuration available (empty configuration present)\")", err.Error())
+		assert.Nil(sm)
+	} else {
+		// for db-mode, the ready endpoint returns 200
```
pmalek, line 80: WDYT about moving this message to assertion so that it's logged when the test fails?

kong/client_test.go
```diff
@@ -34,6 +35,53 @@ func TestKongStatus(T *testing.T) {
...
+
+	client, err := NewTestClientWithOpts(RequestOptions{
+		BaseURL:   String("http://localhost:8001"),
+		StatusURL: String("http://localhost:8100"),
+	}, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
+
+	sm, err := client.Ready(defaultCtx)
+	if err != nil {
```
pmalek, line 75: This can produce false positives when the `err != nil` and we use db-mode. Can we use a different method here to inspect which flavor of Kong we're running against? e.g. inspecting the configuration?

pmalek (changes requested):

sandromodarelli:
closing as https://github.com/Kong/kubernetes-ingress-controller/issues/5068 has been closed as stale

PR #458, opened 2024-08-05
fix: fill config values and defaults in shorthand fields

Prashansa-K:
As of now, deprecated fields are renamed to shorthand_fields in all schemas, as per the conventions. Till now, we do not fill any values for these fields while attempting to fill config records for plugins. This creates a visualisation problem in decK. Everytime, a deck diff or deck sync command is issued, it shows that the deprecated field values are changed and need to be updated, thus running an unnecessary update process each time and also confusing end users.

Check https://github.com/Kong/deck/issues/1251 for details.

This fix attempts to fill defaults for the shorthand_fields, retaining their values, if passed. Also, since shorthand_fields take priority over normal/nested fields in the gateway, if any changes are detected in shorthand_fields, it is backfilled to nested fields as well.

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -208,12 +208,53 @@ func getConfigSchema(schema gjson.Result) (gjson.Result, error) {
 	return schema, fmt.Errorf("no 'config' field found in schema")
 }
 
+func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (interface{}, error) {
+	if len(path) == 0 {
+		return nil, nil
+	}
+
+	pathElement := path[0]
+	value, ok := currentConfigMap[pathElement]
+	if !ok {
+		return nil, fmt.Errorf("key %q not found in map", pathElement)
+	}
+
+	switch v := value.(type) {
+	case map[string]interface{}:
+		// Traversing the map recursively, dissecting the path each time
+		return traverseConfigMap(v, path[1:])
+	default:
+		return v, nil
+	}
+}
+
+func backfillResultConfigMap(res Configuration, path []string, configValue interface{}) error {
+	// Traverse the map to the second-to-last level
+	for i, p := range path {
+		if i == len(path)-1 {
+			// Last element in the path, update the value
+			res[p] = configValue
+			return nil
+		}
+		// Traverse to the next level
+		next, ok := res[p].(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("backward_translation path incorrect")
+		}
+		res = next
+	}
+
+	return nil
+}
+
 func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 	res := config.DeepCopy()
-	value := schema.Get("fields")
+	configFields := schema.Get("fields")
+	// Fetch deprecated fields
+	shortHandFields := schema.Get("shorthand_fields")
 	defaultRecordValue := schema.Get("default")
 
-	value.ForEach(func(_, value gjson.Result) bool {
+	configFields.ForEach(func(_, value gjson.Result) bool {
 		// get the key name
 		ms := value.Map()
 		fname := ""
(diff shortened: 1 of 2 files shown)


kong/utils.go
```diff
@@ -208,12 +208,53 @@ func getConfigSchema(schema gjson.Result) (gjson.Result, error) {
...
+	switch v := value.(type) {
+	case map[string]interface{}:
+		// Traversing the map recursively, dissecting the path each time
+		return traverseConfigMap(v, path[1:])
+	default:
+		return v, nil
+	}
+}
+
+func backfillResultConfigMap(res Configuration, path []string, configValue interface{}) error {
```
pmalek, line 231: Can we add a godoc for this func to explain to the caller how this should be used?

kong/utils.go
```diff
@@ -208,12 +208,53 @@ func getConfigSchema(schema gjson.Result) (gjson.Result, error) {
 	return schema, fmt.Errorf("no 'config' field found in schema")
 }
 
+func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (interface{}, error) {
```
pmalek, line 211: Can we add a godoc for this func to explain to the caller how this should be used?

kong/utils.go
```diff
@@ -327,6 +368,63 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 		return true
 	})
 
+	// Filling defaults for deprecated fields
+	// Required for deck sync/diff inorder
+	// Otherwise, users keep seeing updates in these fields despite of no change
+	shortHandFields.ForEach(func(_, value gjson.Result) bool {
```
pmalek, line 374: I realize we're adding to this func here but it's already rather big and hard to comprehend.

Can we extract the added part to a named func and use it here? IMHO this would greatly enhance the readability of this section of the code.

WDYT?

kong/utils.go
```diff
@@ -208,12 +208,53 @@ func getConfigSchema(schema gjson.Result) (gjson.Result, error) {
...
+	for i, p := range path {
+		if i == len(path)-1 {
+			// Last element in the path, update the value
+			res[p] = configValue
+			return nil
+		}
+		// Traverse to the next level
+		next, ok := res[p].(map[string]interface{})
+		if !ok {
+			return fmt.Errorf("backward_translation path incorrect")
```
pmalek, line 242: Let's include the offending `path` in the error message so that it's easier to debug in case this happens: 

```suggestion
			return fmt.Errorf("backward_translation path %q incorrect", p)
```

kong/utils.go
```diff
@@ -327,6 +368,63 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
...
+			return true
+		}
+
+		var configPathForBackwardTranslation []string
+		for _, value := range backwardTranslation.Array() {
+			configPathForBackwardTranslation = append(configPathForBackwardTranslation, value.Str)
+		}
+
+		if deprecatedFieldValue != nil {
+			_ = backfillResultConfigMap(res, configPathForBackwardTranslation, deprecatedFieldValue)
```
pmalek, line 412: Why not check the error here? 🤔

pmalek (commented):
Some minor comments here and there.

kong/utils.go
Prashansa-K: configFields are added in a similar way. So, I left it as it is to keep both of the iterators similar.
Also, there's nothing much common between the two. So, creating a different function just for one of these seemed out of the place to me.

kong/utils.go
Prashansa-K: Similar reason to why we are not erroring out for other cases. This entire block attempts to fill values/defaults for deprecated fields. Thus, it doesn't seem so crucial to error out.
If we let the thing pass, deck sync or deck diff may show some updates, but the process will still go through.

Adding a comment in the code to explain the same.

pmalek (commented):
Other than the remaining bits, LGTM.

pmalek (approved):

PR #463, opened 2024-08-19
fix(*): avoid setting nil for missing fields

samugi:
when filling defaults, the existing logic sets `nil` on missing fields, but in Kong Gateway explicit `nil`s result in the corresponding fields to be unset, causing problems such as: [KAG-5157](https://konghq.atlassian.net/browse/KAG-5157)

This change **removes** any unset fields that don't have a default from the configuration instead of explicitly setting them to `nil`.

`@reviewers`: could this be a breaking change? I tested it in decK and it seems to work but I'm worried about other projects using this lib. An alternative would be creating another utils function with the new behavior (which I think would require updating at least `deck`, `go-database-reconciler` and probably [kubernetes-ingress-controller](https://github.com/Kong/kubernetes-ingress-controller/blob/09257b8425b3d3654028fa5bbedb7d3f6677079f/internal/dataplane/deckgen/generate.go)), but if we're fairly confident the func is only used while interacting with decK then it *might* be considered safe enough.

[Searching occurrences](https://github.com/search?q=org%3AKong+FillPluginsDefaults&type=code).

For more details see: [KAG-5210](https://konghq.atlassian.net/browse/KAG-5210)

[KAG-5157]: https://konghq.atlassian.net/browse/KAG-5157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KAG-5210]: https://konghq.atlassian.net/browse/KAG-5210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -393,8 +393,8 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 		if value.Exists() {
 			res[fname] = value.Value()
 		} else {
-			// if no default exists, set an explicit nil
-			res[fname] = nil
+			// if no default exists, remove the field
+			delete(res, fname)
 		}
 		return true
 	})
--- a/kong/plugin_service_test.go
+++ b/kong/plugin_service_test.go
@@ -573,7 +573,6 @@ func TestFillPluginDefaults(T *testing.T) {
 				Name:  String("basic-auth"),
 				RunOn: String("test"),
 				Config: Configuration{
-					"anonymous":        nil,
 					"hide_credentials": false,
 					// NOTE: realm has been introduced in 3.6 basic auth schema
 					// https://docs.konghq.com/hub/kong-inc/basic-auth/changelog/#kong-gateway-36x
@@ -594,7 +593,6 @@ func TestFillPluginDefaults(T *testing.T) {
 				Name:  String("basic-auth"),
 				RunOn: String("test"),
 				Config: Configuration{
-					"anonymous":        nil,
 					"hide_credentials": false,
 				},
 				Protocols: []*string{String("grpc"), String("grpcs"), String("http"), String("https")},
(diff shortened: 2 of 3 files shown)


kong/utils.go
```diff
@@ -393,8 +393,8 @@ func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
 		if value.Exists() {
 			res[fname] = value.Value()
 		} else {
-			// if no default exists, set an explicit nil
-			res[fname] = nil
+			// if no default exists, remove the field
+			delete(res, fname)
```
samugi, line 397: this is the actual change

samugi:
not sure why the linter got mad

randmonkey:
> not sure why the linter got mad

https://github.com/Kong/go-kong/actions/runs/10493904848/job/29068947573?pr=463 
Looks like the linter was updated some day but it did not notify us

pmalek:
@samugi https://github.com/Kong/go-kong/pull/464 This should solve the linter issues and align linter version to be the same on CI and locallly

bungle:
For visibility:

I do not fully understand why there is even this `FillPluginsDefaults`. I think it is wrong that the defaults or `null`s are filled to config. If user has `.yaml` file, the tool should not invent any fields in it, and send only those that user has specified, e.g. even when user has specified `field: ~` (aka null), that should be sent to kong, but nothing else. Kong, at least the Lua CP will fill the defaults, and do proper thing.

Now if tool itself needs these to be filled (e.g. for easier diffing or something), it is another question, but that should not affect what is sent to wire to Kong API (provided that Koko API works the same).

This of course means that the tool should also never use `PATCH` or `POST`, just `PUT` and `DELETE` and `GET`.

mheap:
> That said, we can first stop filling nulls (as it fixes the immediate issue of deprecated shorthand precedence), and perhaps later (in other PR) stop filling defaults.

Stop filling default nulls, you mean? Active null should still be accepted

samugi:
> > That said, we can first stop filling nulls (as it fixes the immediate issue of deprecated shorthand precedence), and perhaps later (in other PR) stop filling defaults.
> 
> Stop filling default nulls, you mean? Active null should still be accepted

yes @mheap that is correct, this change is intended to only apply to auto-filled defaults

samugi:
Note: this appears to be breaking is the decK diff output (where now all nulls show up as differences) - probably expected.

samugi:
second attempt: https://github.com/Kong/go-database-reconciler/pull/133

PR #464, opened 2024-08-22
fix(ci): fix linter issues and use deterministic golangci-lint version used both locally and in CI

pmalek:
Use a deterministic version of `golangci-lint` that is the same locally and on CI.


randmonkey (approved):
One step before me :100:

PR #465, opened 2024-08-28
fix(utils): support auto fields

samugi:
When fields marked as `auto` are passed to kong without a value they are auto generated by the gateway. In that case, comparing with "defaults" doesn't work, so in order to compute a correct diff they must be copied from Kong's configuration.

This commit addS a new FillPluginsDefaultsAutoFields utils function that in addition to default fields, adds auto fields to the configuration. It takes an additional parameter: oldPlugin, which represents the plugin as defined in the config from Kong: auto fields are copied from it so that they don't show up in the diff.

[KAG-5210](https://konghq.atlassian.net/browse/KAG-5210)

[KAG-5210]: https://konghq.atlassian.net/browse/KAG-5210?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


kikito (approved):

bungle (approved):

kong/utils.go
```diff
@@ -278,7 +278,7 @@ func backfillResultConfigMap(res Configuration, path []string, configValue inter
 	return nil
 }
 
-func fillConfigRecord(schema gjson.Result, config Configuration) Configuration {
+func fillConfigRecord(schema gjson.Result, config Configuration, kongConfig Configuration) Configuration {
```
randmonkey, line 281: naming: The `kongConfig` seems to infer to the existing configuration. The `config` and `kongConfig` looks a bit confusing to me, could we name it `oldConfig` or `existingConfig` and add comments to explain them?

samugi: done.

Prashansa-K (approved):

randmonkey (approved):

PR #467, opened 2024-09-11
fix(utils): do not fill empty records

samugi:
follow up of: https://github.com/Kong/go-kong/pull/466, prevent unset records to be filled with empty tables: {}

KAG-5380

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -355,7 +355,12 @@ func fillConfigRecord(schema gjson.Result, config Configuration, opts FillRecord
 				fieldConfig = subConfig.(map[string]interface{})
 			}
 			newSubConfig := fillConfigRecord(value.Get(fname), fieldConfig, opts)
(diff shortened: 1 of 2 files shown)


Prashansa-K (approved):

pmalek (commented):
Looks good to me 👍 

I believe this should get a changelog entry, right?

samugi:
oops, looks like some of entries are missing there for the past couple of releases, I think we should open a separate PR to fill those in at once, before releasing, let me know if that sounds good to you @pmalek

pmalek:
Sounds good! Be sure to include the changelog links in the created releases as stated in https://github.com/Kong/go-kong/blob/76af65d56693f60db76520e8bdb7593cdb14acdc/RELEASE.md?plain=1#L5-L7.

pmalek (approved):

PR #472, opened 2024-10-17
fix: use replaced_with to check for deprecated fields

Prashansa-K:
In schemas, shorthand_fields that are deprecated
will use "replaced_with" going forward to convey the new fields that would be used in place of the old ones. Earlier, "translate_backwards" was used for the same.

This change ensures that both of the above mentioned fields can be used to find the new path that leads to the new values for comparison purposes. This will help tools like deck to show diff between syncs.

Fixes [deck #1440](https://github.com/Kong/deck/issues/1404)
Corresponding PR: https://github.com/Kong/kong/pull/13768

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -452,24 +452,24 @@ func fillConfigRecord(schema gjson.Result, config Configuration, opts FillRecord
 		// the defaults for deprecated fields from the already formed default config
 		backwardTranslation := value.Get(fname + ".translate_backwards")
 
-		var replacePath gjson.Result
-		replacedWith := value.Get(fname + ".deprecation.replaced_with")
-		if replacedWith.IsArray() {
-			for _, item := range replacedWith.Array() {
-				if pathArray := item.Get("path"); pathArray.Exists() && pathArray.IsArray() {
-					replacePath = pathArray
-					break
+		if !backwardTranslation.Exists() {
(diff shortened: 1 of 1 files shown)


kong/utils.go
```diff
@@ -452,12 +452,27 @@ func fillConfigRecord(schema gjson.Result, config Configuration, opts FillRecord
...
+		var replacePath gjson.Result
+		replacedWith := value.Get(fname + ".deprecation.replaced_with")
+		if replacedWith.IsArray() {
+			for _, item := range replacedWith.Array() {
+				if pathArray := item.Get("path"); pathArray.Exists() && pathArray.IsArray() {
+					replacePath = pathArray
+					break
+				}
+			}
+		}
```
samugi, line 464: looks like we don't need to run any of this code if `backwardTranslation.Exists()` so we could execute this conditionally to `backwardTranslation` not existing.

I think we may also simplify and just reassign `backwardTranslation = pathArray` here if the item is found, so we wouldn't need the `replacePath` variable.

samugi (commented):
LGTM, left a comment.

samugi (approved):

Prashansa-K:
@pmalek Could you help with an approval here please? We need one from code-owners to merge this PR. Sam has done the code review for this.

randmonkey (approved):

PR #473, opened 2024-10-22
fix(utils): add ClearUnmatchingDeprecations function to align configs

nowNick:
# Summary

When a new plugin configuration is sent that contains information about some deprecated fields in order to produce 
correct diff we need to align the API response with given config.

### Description

This PR aims to align the configs by doing two things:
1. Adding: `ClearUnmatchingDeprecations`
2. Modifying: `fillConfigRecord`

More details can be found in the functions documentation.

### Important changes:

- `nil` values are preserved and not changed when `fillConfigRecord` is run. Previously a config `cluster_max_redirections: null` would be "filled" with `cluster_max_redirections: 5` if `cluster_max_redirections` had a default value of 5 defined. Right now sending explicit  null preserves that. This reverts the change introduced in: https://github.com/Kong/go-kong/pull/309/files#diff-af8350ebad15aeeb0a2561433d1d0907140df8e5b8267a0d0f05bb53d0d53fcbR221-R249
- shorthand values are not backfilled. If a new config contains only new fields `fillConfigRecord` will not try to fill deprecated fields in that config
- if a deprecated field was defined `fillConfigRecord` will not populate the new fields that are linked to that deprecated field

### Related PRs:
- (next one) https://github.com/Kong/go-database-reconciler/pull/145

KAG-5577

### DIFF
--- a/kong/utils.go
+++ b/kong/utils.go
@@ -913,7 +913,7 @@ func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (
 	}
 }
 
-// This function handles the relationship between deprecated and new plugin configuration values.
+// clearUnmatchingDeprecationsForGivenPath handles the relationship between deprecated and new plugin configuration values.
 // We consider the following scenarios:
 //
 // - **Scenario 1**: Both old and new values are present.
--- a/kong/utils_test.go
+++ b/kong/utils_test.go
@@ -2825,11 +2825,13 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
 
 func Test_ClearUnmatchingDeprecationsSimple(t *testing.T) {
 	RunWhenKong(t, ">=3.8.0")
-	assert := assert.New(t)
-
 	client, err := NewTestClient(nil, nil)
-	assert.NoError(err)
-	assert.NotNil(client)
+	require.NoError(t, err)
+	require.NotNil(t, client)
+
+	fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting")
+	require.NoError(t, err)
+	require.NotNil(t, fullSchema)
 
 	tests := []struct {
 		name              string
@@ -2908,9 +2910,6 @@ func Test_ClearUnmatchingDeprecationsSimple(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting")
-			assert.NoError(err)
-			assert.NotNil(fullSchema)
 			require.NoError(t, ClearUnmatchingDeprecations(tc.newPlugin, tc.oldPlugin, fullSchema))
 			if diff := cmp.Diff(tc.oldPlugin.Config, tc.expectedOldPlugin); diff != "" {
 				t.Errorf("unexpected diff:\n%s", diff)
@@ -2921,11 +2920,13 @@ func Test_ClearUnmatchingDeprecationsSimple(t *testing.T) {
 
 func Test_ClearUnmatchingDeprecationsAdvanced(t *testing.T) {
 	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
-	assert := assert.New(t)
-
 	client, err := NewTestClient(nil, nil)
-	assert.NoError(err)
-	assert.NotNil(client)
+	require.NoError(t, err)
+	require.NotNil(t, client)
+
+	fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
+	require.NoError(t, err)
+	require.NotNil(t, fullSchema)
 
 	tests := []struct {
 		name              string
@@ -3139,9 +3140,6 @@ func Test_ClearUnmatchingDeprecationsAdvanced(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
-			assert.NoError(err)
-			assert.NotNil(fullSchema)
 			require.NoError(t, ClearUnmatchingDeprecations(tc.newPlugin, tc.oldPlugin, fullSchema))
 			if diff := cmp.Diff(tc.oldPlugin.Config, tc.expectedOldPlugin); diff != "" {
 				t.Errorf("unexpected diff:\n%s", diff)
@@ -3179,11 +3177,13 @@ func Test_ClearUnmatchingDeprecationsWhenSchemaIsWrong(t *testing.T) {
 
 func Test_ClearUnmatchingDeprecationsWhenNotUpdateEvent(t *testing.T) {
 	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
-	assert := assert.New(t)
-
 	client, err := NewTestClient(nil, nil)
-	assert.NoError(err)
-	assert.NotNil(client)
+	require.NoError(t, err)
+	require.NotNil(t, client)
+
+	fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
+	require.NoError(t, err)
+	require.NotNil(t, fullSchema)
 
 	tests := []struct {
 		name                     string
@@ -3266,9 +3266,6 @@ func Test_ClearUnmatchingDeprecationsWhenNotUpdateEvent(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
-			assert.NoError(err)
-			assert.NotNil(fullSchema)
 			require.NoError(t, ClearUnmatchingDeprecations(tc.newPlugin, tc.oldPlugin, fullSchema))
 			if tc.expectedNewPluginCleared != nil {
 				if diff := cmp.Diff(tc.newPlugin.Config, tc.expectedNewPluginCleared); diff != "" {
@@ -3287,11 +3284,13 @@ func Test_ClearUnmatchingDeprecationsWhenNotUpdateEvent(t *testing.T) {
 
 func Test_ClearUnmatchingDeprecationsWhenNewConfigIsSetAsNil(t *testing.T) {
 	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
-	assert := assert.New(t)
-
 	client, err := NewTestClient(nil, nil)
-	assert.NoError(err)
-	assert.NotNil(client)
+	require.NoError(t, err)
+	require.NotNil(t, client)
+
+	fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
+	require.NoError(t, err)
+	require.NotNil(t, fullSchema)
 
 	tests := []struct {
 		name                     string
@@ -3410,9 +3409,6 @@ func Test_ClearUnmatchingDeprecationsWhenNewConfigIsSetAsNil(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
-			assert.NoError(err)
-			assert.NotNil(fullSchema)
 			require.NoError(t, ClearUnmatchingDeprecations(tc.newPlugin, tc.oldPlugin, fullSchema))
 			if diff := cmp.Diff(tc.newPlugin.Config, tc.expectedNewPluginCleared); diff != "" {
 				t.Errorf("unexpected diff:\n%s", diff)
@@ -3427,11 +3423,13 @@ func Test_ClearUnmatchingDeprecationsWhenNewConfigIsSetAsNil(t *testing.T) {
 
 func Test_ClearUnmatchingDeprecationsWhenNewConfigHasDefaults(t *testing.T) {
 	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
-	assert := assert.New(t)
-
 	client, err := NewTestClient(nil, nil)
-	assert.NoError(err)
-	assert.NotNil(client)
+	require.NoError(t, err)
+	require.NotNil(t, client)
+
+	fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
+	require.NoError(t, err)
+	require.NotNil(t, fullSchema)
 
 	tests := []struct {
 		name                     string
@@ -3550,9 +3548,6 @@ func Test_ClearUnmatchingDeprecationsWhenNewConfigHasDefaults(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
-			assert.NoError(err)
-			assert.NotNil(fullSchema)
 			require.NoError(t, ClearUnmatchingDeprecations(tc.newPlugin, tc.oldPlugin, fullSchema))
 			if diff := cmp.Diff(tc.newPlugin.Config, tc.expectedNewPluginCleared); diff != "" {
 				t.Errorf("unexpected diff:\n%s", diff)


kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
 func FillPluginsDefaultsWithOpts(plugin *Plugin, schema map[string]interface{}, opts FillRecordOptions) error {
 	return fillConfigRecordDefaultsAutoFields(plugin, schema, opts)
 }
+
+func deleteAndCollapseMap(config map[string]interface{}, path []string) {
+	key := path[0]
```
Prashansa-K, line 746: Can we check for the possibility of an empty path and empty map before this to avoid panics during runtime?

kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+	if len(path) == 1 {
+		delete(config, key)
+	} else {
+		if nested, ok := config[key].(map[string]interface{}); ok {
+			deleteAndCollapseMap(nested, path[1:])
+			if len(nested) == 0 {
```
Prashansa-K, line 752: Out of curiosity, when can this case arise?

Prashansa-K (dismissed):

kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+		_, ok := config[key]
+		return ok
+	} else if nested, ok := config[key].(map[string]interface{}); ok {
+		return pathExistsInConfig(nested, path[1:])
+	}
+
+	return false
+}
+
+func clearUnmatchingDeprecationsHelper(
```
Prashansa-K, line 771: Can we add comments over the functions to explain what they do and in which scenarios they may be useful?

nowNick:
Thanks @Prashansa-K for the review. For now I've changed this PR back to draft because I've found a couple of issues.

kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+	shortHandFields := schema.Get("shorthand_fields")
+
+	shortHandFields.ForEach(func(_, value gjson.Result) bool {
+		field := value.Map()
+		for deprecatedFieldName, shorthandFieldConfig := range field {
+			if _, ok := newPluginConfig[deprecatedFieldName]; ok {
+				// deprecatedFieldName is used in new plugin configuration
+				// verify if the fields that this depractedField is replaced with
+				// are also sent in new plugin configuration - if not clear them from old plugin configuration
+				replacements := shorthandFieldConfig.Get("deprecation.replaced_with.#.path")
```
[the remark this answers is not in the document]
nowNick: Yeah I haven't figured out how to deal with backwards compatibility yet 🤔  but that's a good point.

kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
 func FillPluginsDefaultsWithOpts(plugin *Plugin, schema map[string]interface{}, opts FillRecordOptions) error {
 	return fillConfigRecordDefaultsAutoFields(plugin, schema, opts)
 }
+
+func deleteAndCollapseMap(config map[string]interface{}, path []string) {
+	key := path[0]
+	if len(path) == 1 {
+		delete(config, key)
```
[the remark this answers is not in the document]
nowNick: Hmm... does it matter 🤔 ? I mean - it won't enter `else` block and having a `return` right after this line doesn't change anything right? Like there won't be any other instruction either way.


Oh... maybe you mean like:
```go
if condition {
  return do_something()
}
// here would be the else part?
```

?
I'm not sure I fully understand 🤔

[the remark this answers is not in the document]
Prashansa-K: I meant something like this:

```
  if condition {
    // do something
    return
  }
  
  // rest of the code, no else
```

This helps with readability as well.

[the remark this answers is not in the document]
nowNick: Oh... Ok. Honestly I think this is just a matter of taste. For me the current implementation reads better. Is it a _"go" thing_ to add returns at the end of `if / else` blocks? I can switch it if that's how it's usually done in go, otherwise I'd prefer to leave it as it is.

kong/utils.go
nowNick: I wanted to cover a scenario when ultimately the entire nested map needs to collapse. Like lets say we have a structure:
```json
{
  "a": {
    "a1": 1,
    "a2": 2,
    "a3": 3
  },
  "b": {
    "b1": 1,
    "b2": 2,
    "b3": 3,
  }   
}
```

and let's say we have deprecated: `a.a1`, `a.a2`, `a.a3` and `b.b1`, `b.b2`. When collapsing this I'd like to end up with following JSON:

```json
{
  "b": {
    "b3": 3
  }
}
```

kong/utils.go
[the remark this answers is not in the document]
pmalek: +1 to what @Prashansa-K suggests. This way you keep the happy path the least indented one, plus be done with the if statement: there might be other conditions which you'll need to handle below).

kong/utils.go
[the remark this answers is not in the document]
nowNick: Ok so in order to tackle all the quirks of various Gateway versions I decided to leave the old `translate_backwards` code in place. It should cover backwards compatibility issues.

To elaborate a little bit on that: `translate_backwards` was introduced in 3.6. It is visible only in `rate-limiting`, `response-ratelimiting` and `acme` plugins. Those deprecations contain only simple renaming. Thus the previous code that copied the values between fields is sufficient enough to cover those scenarios.

Now, `replaced_with` was introduced in 3.8 and it contains more sophisticated information as it covers more sophisticated scenarios of not only renames. So for everything 3.6 >= x > 3.8 we should be good with the code that's already there and for everything above 3.8 we should be good with code that's proposed here. We only need to clear unmatching deprecations on the fields that were introduced in 3.8.

kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+					clearUnmatchingDeprecationsHelper(nestedNewPluginConfig, nestedOldPluginConfig, &fieldConfig)
+				}
+			}
+		}
+
+		return true
+	})
+}
+
+func ClearUnmatchingDeprecations(newPlugin *Plugin, oldPlugin *Plugin, schema map[string]interface{}) error {
```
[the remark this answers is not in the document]
nowNick: I've pointed the corresponding PR to this one so the tests there are passing as well.

kong/utils.go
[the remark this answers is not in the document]
nowNick: Alright, I've changed the code to requested style :)

kong/utils.go
nowNick: Done ✅

kong/utils.go
nowNick: Done ✅

kong/utils.go
```diff
@@ -741,3 +741,119 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+					for i, pathSegment := range value.Array() {
+						replacementPathAsStringArray[i] = pathSegment.String()
+					}
+
+					// We know that deprecated value is defined in new config and we also have information
+					// on how this deprecated value is replaced. If the new plugin configuration contains
+					// both old and new values we don't need to adjust old plugin configuration.
+					// However if the new key is missing in new plugin configuration then we need to
+					// delete it from old plugin configuration in order for them to match.
+					if !pathExistsInConfig(newPluginConfig, replacementPathAsStringArray) {
```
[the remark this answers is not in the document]
nowNick: Done ✅

nowNick:
Ok the corresponding PR is green and I think I've tackled the edge cases I've discovered (in the tests in this PR and the other one: https://github.com/Kong/go-database-reconciler/pull/145 ). I think it's ready to be reviewed again.

kong/utils.go
```diff
@@ -741,3 +741,147 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+
+					// We know that deprecated value is defined in new config and we also have information
+					// on how this deprecated value is replaced. If the new plugin configuration contains
+					// both old and new values we don't need to adjust old plugin configuration.
+					// However if the new key is missing in new plugin configuration then we need to
+					// delete it from old plugin configuration in order for them to match.
+					// --
+					// There is also a possibility that the new field exists in new config but only because
+					// it was filled with `null` value due to decK logic of filling empty keys. In that case
+					// if the deprecated value is different from nil then we need to clear it both in new and old config.
```
samugi, line 814: I think a bullet-list with all the different scenarios, and what we do for each, would be more helpful as a description of the code below

kong/utils.go
```diff
@@ -741,3 +741,147 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+	// Fetch deprecated fields
+	shortHandFields := schema.Get("shorthand_fields")
+
+	shortHandFields.ForEach(func(_, value gjson.Result) bool {
+		field := value.Map()
+		for deprecatedFieldName, shorthandFieldConfig := range field {
+			if deprecatedFieldValue, ok := newPluginConfig[deprecatedFieldName]; ok {
+				// deprecatedFieldName is used in new plugin configuration
+				// verify if the fields that this depractedField is replaced with
+				// are also sent in new plugin configuration - if not clear them from old plugin configuration
```
samugi, line 798: this comment does not seem to refer to the code right below it, and the same logic is explained again below, I would remove these 2 lines of comment.

kong/utils.go
```diff
@@ -741,3 +741,147 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+					// on how this deprecated value is replaced. If the new plugin configuration contains
+					// both old and new values we don't need to adjust old plugin configuration.
+					// However if the new key is missing in new plugin configuration then we need to
+					// delete it from old plugin configuration in order for them to match.
+					// --
+					// There is also a possibility that the new field exists in new config but only because
+					// it was filled with `null` value due to decK logic of filling empty keys. In that case
+					// if the deprecated value is different from nil then we need to clear it both in new and old config.
+					acceptNullValue := deprecatedFieldValue == nil
+					newPluginFieldValue, _ := traverseConfigMap(newPluginConfig, replacementPathAsStringArray)
```
samugi, line 816: should we call it newPluginField**New**Value? :D

samugi (approved):

kong/utils.go
nowNick: I'll see if I can come up with a better name :D

GGabriele (approved):

kong/utils.go
nowNick: Fixed ✅

kong/utils.go
nowNick: Well... the best I could come up with was: `newPluginNewFieldValue`. LLMs didn't help here either. 🙁

kong/utils.go
nowNick: Fixed ✅

nowNick:
There's one edge-case left that I'm working on. I have a POC ready and I'll refactor, add to this PR and verify if it fixes the issue found in: https://github.com/Kong/go-database-reconciler/pull/145#issuecomment-2450336564
(FYI: @Prashansa-K )

pmalek (commented):
Minor comments on Go style etc.

kong/utils.go
```diff
@@ -741,3 +865,204 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+// It does so by removing deprecated or new fields in "oldPlugin" that were not defined in "oldPlugin".
+// Furthermore it'll remove new field from newPlugin when it's value is nil and the corresponding deprecated
+// value is not nil (in that case we can be sure that "oldPlugin" contains
+// specific value for both new and old fields).
+func ClearUnmatchingDeprecations(newPlugin *Plugin, oldPlugin *Plugin, schema map[string]interface{}) error {
+	jsonb, err := json.Marshal(&schema)
+	if err != nil {
+		return err
+	}
+	gjsonSchema := gjson.ParseBytes((jsonb))
```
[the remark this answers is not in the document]
nowNick: Fixed ✅

kong/utils.go
```diff
@@ -249,33 +249,206 @@ func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (
...
 		}
-		res = next
+
+		return paths
 	}
 
-	return nil
+	backwardTranslation := shorthandField.Get("translate_backwards")
+	if !backwardTranslation.Exists() {
+		return make([][]string, 0)
```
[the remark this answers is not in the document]
nowNick: Fixed ✅  - I'm returning `nil` and then checking if `nil` was returned. 

From style perspective I'd prefer to keep it more monadic (empty array) but as you pointed out that would hinder performance & memory a little bit... plus, I'm not sure if go is that much into functional programming so returning nil is probably more idiomatic to the language.

kong/utils.go
```diff
@@ -249,33 +249,206 @@ func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (
...
+	}
+	paths := make([][]string, 1)
+	paths[0] = readStringArray(backwardTranslation)
```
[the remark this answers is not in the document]
nowNick: Fixed ✅

kong/utils.go
```diff
@@ -249,33 +249,206 @@ func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (
...
+// Example:
+//
+//		gjsonArray := gjson.Parse(`["apple", "banana", "cherry"]`)
+//		result := readStringArray(gjsonArray)
+//		fmt.Println(result)
+//		// Output:
+//	 	// ["apple", "banana", "cherry"]
+func readStringArray(gjsonArray gjson.Result) []string {
+	result := make([]string, len(gjsonArray.Array()))
+	for i, pathSegment := range gjsonArray.Array() {
```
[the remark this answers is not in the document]
nowNick: Fixed ✅

Prashansa-K:
I tested these changes locally with an openid plugin (deck file added [here](https://github.com/Kong/go-database-reconciler/pull/145#issuecomment-2450336564))and noticed the following:

On gateway 3.8.1.0
No unnecessary diffs come up, except 
```
+    "session_compressor": "none"
+    "session_cookie_maxsize": 4000
+    "session_cookie_renew": 600
+    "session_strategy": "default"
```
These fields don't have any translate_backwards or replaced_with property in the schema. So, not applicable with this change.

However, on Konnect, I noticed some diffs:
```
-    "session_redis_cluster_max_redirections": 5,
+    "session_redis_cluster_max_redirections": null,

-    "session_redis_connect_timeout": 2000,
+    "session_redis_connect_timeout": null,
    
-    "session_redis_read_timeout": 2000,
+    "session_redis_read_timeout": null,
-    "session_redis_send_timeout": 2000,
+    "session_redis_send_timeout": null,
```

decK and reconciler logic remains same for Konnect as far as I know. Is there anything different schema-wise for Konnect that could be causing this?

kong/utils.go
```diff
@@ -249,33 +249,206 @@ func traverseConfigMap(currentConfigMap map[string]interface{}, path []string) (
...
+//
+//	map[string][][]string: A map of deprecated field names to replacement paths, with invalid fields removed.
+func processDeprecatedFields(shorthandFields gjson.Result, config Configuration) map[string][][]string {
+	deprecatedFieldsWithReplacements := buildDeprecatedFieldWithReplacementsMap(shorthandFields)
+	filterValidShorthands(deprecatedFieldsWithReplacements, config)
+
+	return deprecatedFieldsWithReplacements
+}
+
+// filterValidShorthands removes any shorthand fields from the map that are not found in the given config.
```
samugi, line 404: what does it mean when a shorthand field is in the map but not in the config? Could you give an example when/why this would be expected to happen?

nowNick:
@Prashansa-K it seems that it's a well known Koko issue. The reason why you see these diffs in Konnect is that Lua-CP and Koko treat `null` values differently. LuaCP preserves them but Koko falls back to default value if defined.

kong/utils.go
samugi: got it thanks for clarifying!

kong/utils.go
```diff
@@ -741,3 +824,247 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+	// Recursively walk through configuration to clear any nested unmatching deprecations.
+	traverseConfigurationsAndExecute(
+		newPluginConfig,
+		oldPluginConfig,
+		schema,
+		clearUnmatchingDeprecationsHelper,
+	)
+}
+
+// ClearUnmatchingDeprecations is a function that go through a pair of
```
pmalek, line 1047:
```suggestion
// ClearUnmatchingDeprecations is a function that goes through a pair of
```

kong/utils.go
```diff
@@ -741,3 +824,247 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+// configurations: newPlugin and oldPlugin, and by using schema it makes sure those two configurations
+// are aligned.
+// It does so by removing deprecated or new fields in "oldPlugin" that were not defined in "oldPlugin".
```
pmalek, line 1050: > removing deprecated or new fields in "oldPlugin" that were not defined in "oldPlugin"

This is somewhat unclear. This removes things from `oldPlugin` which were not defined in `oldPlugin` 🤔 ?

kong/utils.go
```diff
@@ -741,3 +824,247 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+// Furthermore it'll remove new field from newPlugin when it's value is nil and the corresponding deprecated
+// value is not nil (in that case we can be sure that "oldPlugin" contains
+// specific value for both new and old fields).
```
pmalek, line 1053: > in that case we can be sure that "oldPlugin" contains specific value for both new and old fields

How can we be sure about this?

kong/utils.go
```diff
@@ -741,3 +824,247 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+				}
+			}
+		}
+
+		return true
+	})
+}
+
+// same as ClearUnmatchingDeprecations but this function below is adjusted for recursive use (which is required
+// when the schema contains nested records and they can have nested shorthands).
```
pmalek, line 1030:
```suggestion
// clearUnmatchingDeprecationsHelper is simmilar to ClearUnmatchingDeprecations but it is adjusted for recursive use (which is required
// when the schema contains nested records and they can have nested shorthands).
```

kong/utils_test.go
```diff
@@ -2822,3 +3142,710 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
 		})
 	}
 }
+
+func Test_ClearUnmatchingDeprecationsSimple(t *testing.T) {
+	tests := []struct {
+		name                     string
+		newPlugin                *Plugin
+		oldPlugin                *Plugin
+		expectedOldPluginCleared Configuration
```
pmalek, line 3151: Nit: I'd rename this to 

```suggestion
		expectedOldPlugin Configuration
```

the `cleared` suffix implies that this is something that was cleared (removed) from `oldPlugin` rather then being the thing that we expect as is.

pmalek (commented):
Some minor comments. Other than that this looks solid. 

One thought though: could we add an integration test that checks against the schema we receive from Gateway? Perhaps we can use https://github.com/Kong/go-kong/blob/5236d86ec5d61693d79db2b85bf347a9060b9247/kong/plugin_service.go#L64 to get the schema and then use that in the test? We can use `RunWhenKong()` to constrain the versions against which we run these tests.

samugi (approved):
I verified the "important changes" described in the summary, LGTM

GGabriele (approved):

kong/utils.go
nowNick: Fixed ✅

kong/utils.go
nowNick: Fixed ✅

nowNick:
@pmalek I've made the changes according to your comments and I've switched to calling Kong API (didn't know we can do this from `go-kong`). Does it look better now?

kong/utils_test.go
```diff
@@ -2822,3 +2822,745 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
...
+	}
+}
+
+func Test_ClearUnmatchingDeprecationsWhenNewConfigIsSetAsNil(t *testing.T) {
+	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 3294:
```suggestion
	require.NoError(err)
	require.NotNil(client)
```

kong/utils.go
```diff
@@ -741,3 +824,281 @@ func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
...
+	switch v := value.(type) {
+	case map[string]interface{}:
+		// Traversing the map recursively, dissecting the path each time
+		return traverseConfigMap(v, path[1:])
+	default:
+		return v, ok
+	}
+}
+
+// This function handles the relationship between deprecated and new plugin configuration values.
```
pmalek, line 916:
```suggestion
// clearUnmatchingDeprecationsForGivenPath handles the relationship between deprecated and new plugin configuration values.
```

nowNick: fixed ✅

kong/utils_test.go
nowNick: fixed ✅

kong/utils_test.go
```diff
@@ -2822,3 +2822,745 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
...
+	}
+}
+
+func Test_ClearUnmatchingDeprecationsWhenNewConfigHasDefaults(t *testing.T) {
+	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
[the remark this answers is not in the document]
nowNick: fixed ✅

kong/utils_test.go
```diff
@@ -2822,3 +2822,745 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
...
+				},
+			},
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			fullSchema, err := client.Schemas.Get(defaultCtx, "plugins/rate-limiting-advanced")
+			assert.NoError(err)
+			assert.NotNil(fullSchema)
```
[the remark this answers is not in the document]
nowNick: fixed ✅  - switched to framework function: `require.NoError(t, ...)`

kong/utils_test.go
```diff
@@ -2822,3 +2822,745 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
...
+	}
+}
+
+func Test_ClearUnmatchingDeprecationsAdvanced(t *testing.T) {
+	RunWhenEnterprise(t, ">=3.8.0", RequiredFeatures{})
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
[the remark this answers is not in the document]
nowNick: fixed ✅

kong/utils_test.go
```diff
@@ -2822,3 +2822,745 @@ func Test_FillPluginsDefaults_NonEmptyDefaultArrayField(t *testing.T) {
...
 	}
 }
+
+func Test_ClearUnmatchingDeprecationsSimple(t *testing.T) {
+	RunWhenKong(t, ">=3.8.0")
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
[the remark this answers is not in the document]
nowNick: fixed ✅

PR #485, opened 2024-11-27
fix: consumer-group scoped plugins not showing tags

Prashansa-K:
On running deck file render command, tags were
coming up for global and service/route scoped
plugins. However, for consumer-group scoped
plugins, the tags did not populate. This was because the underlying struct did not have tags in the first place.

This change updates the underlying struct for
consumer-group plugins by adding tags to them
as well. This will ensure that tags show up
during deck file render.
For: https://github.com/Kong/deck/issues/1417

PR dependent on this: https://github.com/Kong/go-database-reconciler/pull/151


pmalek (commented):
Looks good to me. Do you mind adding some tests in https://github.com/Kong/go-kong/blob/21ef700df085a331e783cddbd2c14d2df3422a01/kong/consumer_group_test.go for this?

Prashansa-K:
> Looks good to me. Do you mind adding some tests in https://github.com/Kong/go-kong/blob/21ef700df085a331e783cddbd2c14d2df3422a01/kong/consumer_group_test.go for this?

@pmalek There are no methods to do CRUD operations on a ConsumerGroupPlugin or a ConsumerGroupObject. If I create a normal plugin associated with a ConsumerGroup and fetch the ConsumerGroupObject via the group id, the plugins are nil. I tested in newer gateway versions as well with 2.8.  I committed this - https://github.com/Kong/go-kong/pull/485/commits/5f2a96d82adbb6f4f51107b5c1bef3e955276885 - to test out but it fails as well. Tried multiple other iterations locally with 3.4, 3.8, 2.8
Can you suggest a way to test this?

I require this because GDR still uses `kong.ConsumerGroupPlugin` in some places.

pmalek:
> > Looks good to me. Do you mind adding some tests in https://github.com/Kong/go-kong/blob/21ef700df085a331e783cddbd2c14d2df3422a01/kong/consumer_group_test.go for this?
> 
> @pmalek There are no methods to do CRUD operations on a ConsumerGroupPlugin or a ConsumerGroupObject. If I create a normal plugin associated with a ConsumerGroup and fetch the ConsumerGroupObject via the group id, the plugins are nil. I tested in newer gateway versions as well with 2.8. I committed this - [5f2a96d](https://github.com/Kong/go-kong/commit/5f2a96d82adbb6f4f51107b5c1bef3e955276885) - to test out but it fails as well. Tried multiple other iterations locally with 3.4, 3.8, 2.8 Can you suggest a way to test this?
> 
> I require this because GDR still uses `kong.ConsumerGroupPlugin` in some places.

Ah I see. `ConsumerGroupPlugin` is only used in deck and in `FillEntityDefaults` in go-kong. I guess that in this case it'd be enough to add tests for this in deck's PR which adds usage of this field 👍

pmalek (approved):
Please add a changelog entry for this, otherwise 👍

PR #487, opened 2024-12-06
feat: do not list consumers in a GET Consumer Group call

Prashansa-K:
Gateways before 3.9 return a list of consumers
along with consumer_group information, when
performing a GET call on a specific consumer_group. This can become slow in case a consumer_group has
a lots of consumers. A separate endpoint and relevant constructs already exist to fetch consumers for a
particular group. Thus, we are trying to not list
consumers while fetching a consumer_group by passing a query parameter in the API call. Gateway 3.9 supports this.

This will make decK faster for cases where a consumer- group holds thousands of consumers.
For: https://github.com/Kong/deck/issues/1457

### DIFF
--- a/kong/consumer_group_service.go
+++ b/kong/consumer_group_service.go
@@ -63,7 +63,7 @@ func (s *ConsumerGroupService) Get(ctx context.Context,
 		return nil, fmt.Errorf("nameOrID cannot be nil for Get operation")
 	}
 
-	endpoint := fmt.Sprintf("/consumer_groups/%v", *nameOrID)
+	endpoint := fmt.Sprintf("/consumer_groups/%v?list_consumers=false", *nameOrID)
 	req, err := s.client.NewRequest("GET", endpoint, nil, nil)
 	if err != nil {
 		return nil, err
--- a/kong/consumer_group_test.go
+++ b/kong/consumer_group_test.go
@@ -242,3 +242,106 @@ func compareConsumerGroups(expected, actual []*ConsumerGroup) bool {
 
 	return (compareSlices(expectedNames, actualNames))
 }
+
+func TestConsumerGroupGetEndpointPreGW39(t *testing.T) {
+	RunWhenEnterprise(t, ">=2.7.0 <3.9.0", RequiredFeatures{})
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	require.NoError(t, err)
+	assert.NotNil(client)
+
+	// create consumer
+	consumer := &Consumer{
+		Username: String("foo"),
+	}
+	createdConsumer, err := client.Consumers.Create(defaultCtx, consumer)
+	require.NoError(t, err)
+	assert.NotNil(createdConsumer)
+	t.Cleanup(func() {
+		err = client.Consumers.Delete(defaultCtx, createdConsumer.ID)
+		assert.NoError(err)
+	})
+
+	// create a consumer-group
+	cg := &ConsumerGroup{
+		Name: String("foo-group"),
+	}
+
+	createdConsumerGroup, err := client.ConsumerGroups.Create(defaultCtx, cg)
+	require.NoError(t, err)
+	assert.NotNil(createdConsumerGroup)
+	t.Cleanup(func() {
+		err = client.ConsumerGroups.Delete(defaultCtx, createdConsumerGroup.ID)
+		assert.NoError(err)
+	})
+
+	// add consumer to consumer group
+	var response *ConsumerGroupObject
+	response, err = client.ConsumerGroupConsumers.Create(defaultCtx, cg.Name, consumer.Username)
+	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


kong/consumer_group_test.go
```diff
@@ -242,3 +242,106 @@ func compareConsumerGroups(expected, actual []*ConsumerGroup) bool {
 
 	return (compareSlices(expectedNames, actualNames))
 }
+
+func TestConsumerGroupGetEndpointPreGW39(t *testing.T) {
+	RunWhenEnterprise(t, ">=2.7.0 <3.9.0", RequiredFeatures{})
+	assert := assert.New(t)
+
+	client, err := NewTestClient(nil, nil)
+	require.NoError(t, err)
```
pmalek, line 251: Nothing major but this is inconsistent here: we create an `assert` object but for `require` we use the top level funcs from the package.

kong/consumer_group_test.go
```diff
@@ -242,3 +242,106 @@ func compareConsumerGroups(expected, actual []*ConsumerGroup) bool {
...
+	consumer := &Consumer{
+		Username: String("foo"),
+	}
+	createdConsumer, err := client.Consumers.Create(defaultCtx, consumer)
+	require.NoError(t, err)
+	assert.NotNil(createdConsumer)
+	t.Cleanup(func() {
+		err = client.Consumers.Delete(defaultCtx, createdConsumer.ID)
+		assert.NoError(err)
+	})
```
pmalek, line 264: 👍 

nit: if `createdConsumer` is `nil`, the assert above can fail (without stopping the test) and cleanup will panic on `createdConsumer.ID`

kong/consumer_group_test.go
```diff
@@ -242,3 +242,106 @@ func compareConsumerGroups(expected, actual []*ConsumerGroup) bool {
...
+		Name: String("foo-group"),
+	}
+
+	createdConsumerGroup, err := client.ConsumerGroups.Create(defaultCtx, cg)
+	require.NoError(t, err)
+	assert.NotNil(createdConsumerGroup)
+	t.Cleanup(func() {
+		err = client.ConsumerGroups.Delete(defaultCtx, createdConsumerGroup.ID)
+		assert.NoError(err)
+	})
```
pmalek, line 277: 👍

kong/consumer_group_test.go
```diff
@@ -242,3 +242,106 @@ func compareConsumerGroups(expected, actual []*ConsumerGroup) bool {
...
+	// create consumer
+	consumer := &Consumer{
+		Username: String("foo"),
+	}
+	createdConsumer, err := client.Consumers.Create(defaultCtx, consumer)
+	require.NoError(t, err)
+	assert.NotNil(createdConsumer)
+	t.Cleanup(func() {
+		err = client.Consumers.Delete(defaultCtx, createdConsumer.ID)
+		assert.NoError(err)
```
pmalek, line 263: Might as well do these, WDYT?

```suggestion
		assert.NoError(client.Consumers.Delete(defaultCtx, createdConsumer.ID))
```

pmalek (commented):
Please make sure to add a changelog to this. Otherwise just minor nits.

kong/consumer_group_test.go
```diff
@@ -242,3 +242,112 @@ func compareConsumerGroups(expected, actual []*ConsumerGroup) bool {
...
+		Username: String("foo"),
+	}
+	createdConsumer, err := client.Consumers.Create(defaultCtx, consumer)
+	require.NoError(err)
+	assert.NotNil(createdConsumer)
+	t.Cleanup(func() {
+		if createdConsumer != nil {
+			assert.NoError(client.Consumers.Delete(defaultCtx, createdConsumer.ID))
+		}
+	})
```
pmalek, line 266: Nit: potential panics when `createdConsumer` is `nil` can still happen but that's not something that we have to worry about that much.

pmalek (approved):

kong/consumer_group_service.go
```diff
@@ -63,7 +63,7 @@ func (s *ConsumerGroupService) Get(ctx context.Context,
 		return nil, fmt.Errorf("nameOrID cannot be nil for Get operation")
 	}
 
-	endpoint := fmt.Sprintf("/consumer_groups/%v", *nameOrID)
+	endpoint := fmt.Sprintf("/consumer_groups/%v?list_consumers=false", *nameOrID)
```
pmalek, line 66: So according to https://github.com/Kong/go-kong/pull/488/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR82-R83 this is only available in GW 3.9+. What will happen when we provide these query params for older versions?

Prashansa-K: The query param is ignored in gw <3.9
I have added a test case for the same too to ensure that things don't break with older versions.

PR #490, opened 2024-12-10
feat: client apis to interact with the konnect applications

reversearrow:
Add Client APIs to interact with the Konnect Applications.


pmalek (dismissed):
Just making a note in this PR to possibly consider using https://github.com/Kong/sdk-konnect-go/ for this.

While at this moment the SDK does not support work with Konnect Applications but only with Control Planes and their configuration we can easily extend that by adding applications related specification to OpenAPI spec https://github.com/Kong/sdk-konnect-go/blob/main/openapi.yaml

I'd be happy to discuss this.

kong/konnect_application.go
```diff
@@ -0,0 +1,27 @@
...
+	ClientID           string              `json:"client_id"`
+	ConsumerGroups     []string            `json:"consumer_groups"`
+	Scopes             []string            `json:"scopes"`
+	AuthStrategyID     *string             `json:"auth_strategy_id"`
+	ApplicationContext *ApplicationContext `json:"application_context"`
+	ExhaustedScopes    []string            `json:"exhausted_scopes"`
+	Tags               *[]string           `json:"tags"`
+}
+
+// ApplicationContext reprensents the application context inside the
```
pmalek, line 19:
```suggestion
// ApplicationContext represents the application context inside the
```

kong/konnect_application_service.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+var _ AbstractKonnectApplicationService = &KonnectApplicationService{}
+
+// AbstractACLService handles consumer ACL groups in Kong.
```
pmalek, line 11:
```suggestion
// AbstractKonnectApplicationService handles Konnect applications in Kong.
```

kong/konnect_application_service_test.go
```diff
@@ -0,0 +1,105 @@
...
+	require.NoError(err)
+	require.NotNil(createResponse)
+	require.Equal(createResponse.ClientID, clientID)
+	require.Equal(createResponse.CreatedAt, createdAt)
+	require.Equal(createResponse.ConsumerGroups, consumerGroup)
+	require.Equal(createResponse.Scopes, scopes)
+	require.Equal(createResponse.ExhaustedScopes, exhaustedScopes)
+
+	err = client.KonnectApplication.Delete(defaultCtx, createResponse.ID)
+	require.NoError(err)
```
pmalek, line 57: Just to be sure that the entity gets deleted when one of the assertions fail

```suggestion
	createResponse, err := client.KonnectApplication.Create(defaultCtx, kaa)
	require.NoError(err)
	require.NotNil(createResponse)
	
	t.Cleanup(func() {
		assert.NoError(client.KonnectApplication.Delete(context.Background(), createResponse.ID))
	})
	
	require.Equal(createResponse.ClientID, clientID)
	require.Equal(createResponse.CreatedAt, createdAt)
	require.Equal(createResponse.ConsumerGroups, consumerGroup)
	require.Equal(createResponse.Scopes, scopes)
	require.Equal(createResponse.ExhaustedScopes, exhaustedScopes)
```

kong/konnect_application_service_test.go
```diff
@@ -0,0 +1,105 @@
...
+		CreatedAt: createdAt,
+	}
+
+	for i := 0; i < expectedKonnectApplications; i++ {
+		clientID := uuid.NewString()
+		kaa.ID = &clientID
+		kaa.ClientID = clientID
+		createResponse, err := client.KonnectApplication.Create(defaultCtx, kaa)
+		require.NoError(err)
+		require.NotNil(createResponse)
```
pmalek, line 99:
```suggestion
		createResponse, err := client.KonnectApplication.Create(defaultCtx, kaa)
		require.NoError(err)
		require.NotNil(createResponse)
		
		t.Cleanup(func() {
			assert.NoError(client.KonnectApplication.Delete(context.Background(), createResponse.ID))
	})
```

kong/konnect_application_service_test.go
pmalek, line 1: Can we also add a test for `List()` specifying the list options?

pmalek (changes requested):
Please make sure to add a changelog entry.

Other than those minor nits this looks good 👍

kong/konnect_application_service_test.go
pmalek: I can see there are several options there

https://github.com/Kong/go-kong/blob/a3495bc5e683b387b7716659dae41c05fa029fbc/kong/list.go#L11-L17

I'd suggest testing `Tags` and `Size`. `ListAll()` doesn't do any filtering. I believe it would be good to test it here. WDYT?

kong/konnect_application_service.go
```diff
@@ -0,0 +1,99 @@
...
+	if isEmptyString(ID) {
+		return fmt.Errorf("ID cannot be nil for Delete operation")
+	}
+
+	req, err := k.client.NewRequest("DELETE", fmt.Sprintf("/konnect_applications/%s", *ID), nil, nil)
+	if err != nil {
+		return err
+	}
+
+	_, err = k.client.Do(ctx, req, nil)
```
pmalek, line 97: I just realized: what's the expected status code for delete? I believe we should check that it's as expected and if it's not then return an error. Otherwise the caller would expect that the object's deleted where in fact we silently ignored the error.

kong/konnect_application_service_test.go
```diff
@@ -0,0 +1,203 @@
...
+func TestKonnectApplicationService_Create(T *testing.T) {
+	RunWhenDBMode(T, "postgres")
+	RunWhenEnterprise(T, ">=3.6.0", RequiredFeatures{})
+
+	assert := assert.New(T)
+	require := require.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	assert.NoError(err)
+	assert.NotNil(client)
```
pmalek, line 23: Just to prevent test going any further when this fails:

```suggestion
	require.NoError(err)
	require.NotNil(client)
```

pmalek (approved):

PR #494, opened 2025-01-08
fix: added new method to fetch a consumer-group without consumers.

Prashansa-K:
Reverted the existing `Get` method to keep listing consumers and added a new one that uses the query
parameter `list_consumers=false` to skip listing
consumers for performance reasons.
This is done so as to fix dump errors coming in
deck when using gw 3.9+ where consumer-groups
do not show the associated consumers. We will
push the performant option behind a flag.

For: https://github.com/Kong/deck/issues/1483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [v0.63.0](#v0630)
 - [v0.62.0](#v0620)
 - [v0.61.0](#v0610)
 - [v0.60.0](#v0600)
@@ -76,9 +77,9 @@
 - [0.2.0](#020)
 - [0.1.0](#010)
 
-## Unreleased
+## [v0.63.0]
 
-> Release date: TBA
+> Release date: 2025/01/10
 
 - Added `GetWithNoConsumers` function to AbstractConsumerGroupService
   to allow the fetching a consumer-group without listing its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1001,6 +1001,8 @@ authentication credentials in Kong.
   releases of Kong since every release of Kong is introducing breaking changes
   to the Admin API.
 
+[v0.61.0]: https://github.com/Kong/go-kong/compare/v0.62.0...v0.63.0
+[v0.61.0]: https://github.com/Kong/go-kong/compare/v0.61.0...v0.62.0
 [v0.61.0]: https://github.com/Kong/go-kong/compare/v0.60.0...v0.61.0
 [v0.60.0]: https://github.com/Kong/go-kong/compare/v0.59.1...v0.60.0
 [v0.59.1]: https://github.com/Kong/go-kong/compare/v0.59.0...v0.59.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Table of Contents
 
+- [v0.62.0](#v0620)
 - [v0.61.0](#v0610)
 - [v0.60.0](#v0600)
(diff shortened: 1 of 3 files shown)


Prashansa-K:
@pmalek Could you please help with this PR review?
I need to unblock the use of `deck gateway dump` for GW 3.9+

CHANGELOG.md
```diff
@@ -75,6 +76,17 @@
 - [0.2.0](#020)
 - [0.1.0](#010)
 
+## Unreleased
```
Prashansa-K, line 79: If the changes and the changelog details look fine, I will update the version and release date in the PR itself.

kong/consumer_group_test.go
```diff
@@ -348,6 +348,15 @@ func TestConsumerGroupGetEndpointPostGW39(t *testing.T) {
...
+	assert.NotNil(consumerGroupFromKong.Consumers)
+	assert.Len(consumerGroupFromKong.Consumers, 1)
+
+	// Check GetWithConsumers
+	consumerGroupFromKong, err = client.ConsumerGroups.GetWithNoConsumers(defaultCtx, createdConsumerGroup.ID)
+	require.NoError(err)
+	assert.NotNil(consumerGroupFromKong)
+	assert.Equal(consumerGroupFromKong.ConsumerGroup.ID, createdConsumerGroup.ID)
+	// Consumers are not listed
 	assert.Nil(consumerGroupFromKong.Consumers)
```
czeslavo, line 361: For readability, I'd suggest putting assertions regarding `GetWithNoConsumers` and `Get` methods into separate test cases. Also, instead of comments, it might be better to pass a message to the assertion. This way, we get to have the same context from the reader's perspective and additionally get a nice message if the test fails. Like so:
```suggestion
	t.Run("GetWithNoConsumers", func(t *testing.T) {
		consumerGroupFromKong, err := client.ConsumerGroups.GetWithNoConsumers(defaultCtx, createdConsumerGroup.ID)
		require.NoError(err)
		assert.NotNil(consumerGroupFromKong)
		assert.Equal(consumerGroupFromKong.ConsumerGroup.ID, createdConsumerGroup.ID)
		assert.Nil(consumerGroupFromKong.Consumers, "consumer should not be listed")
	})
```

CHANGELOG.md
pmalek: I believe the changes look good. When @czeslavo's comment is covered I believe we can merge 👍

Prashansa-K:
@pmalek @czeslavo I have incorporated the requested changes.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Table of Contents
 
+- [v0.63.0](#v0630)
+- [v0.62.0](#v0620)
```
pmalek, line 4: We also need to define these links. This is currently being done at the very bottom of this file.

CHANGELOG.md
```diff
@@ -1001,6 +1001,8 @@ authentication credentials in Kong.
   releases of Kong since every release of Kong is introducing breaking changes
   to the Admin API.
 
+[v0.61.0]: https://github.com/Kong/go-kong/compare/v0.62.0...v0.63.0
+[v0.61.0]: https://github.com/Kong/go-kong/compare/v0.61.0...v0.62.0
```
pmalek, line 1005:
```suggestion
[v0.63.0]: https://github.com/Kong/go-kong/compare/v0.62.0...v0.63.0
[v0.62.0]: https://github.com/Kong/go-kong/compare/v0.61.0...v0.62.0
```

pmalek (approved):

PR #507, opened 2025-02-11
feat: added client apis for partials entity and plugin-partial linking

Prashansa-K:
Gateway 3.10 is introducing a new core-entity "partials", that can
hold some part of plugin configuration in them for easy
cross-reference without duplication across multiple plugins.
For example, a partial "redis-ee" can contain redis configuration
that can be used in multiple RLA plugins.

This PR adds the necessary client APIs required to perform
CRUD operations on partials. It also adds necessary changes in
plugin struct and plugin service to allow partials to work.

Reference Docs (internal):
- [Engineering Brief](https://docs.google.com/document/d/1vwUS5aCHXOjE5jYaysU8eJ59da5OdskoNd58pJGLO2g/edit?tab=t.0)
- [Product Brief](https://konghq.atlassian.net/wiki/spaces/PM/pages/3250290739/Product+Brief+Referenceable+Redis+Configuration)

Some pre-release deck binaries are present here that use the changes in this PR:
https://kongstrong.slack.com/archives/C0663589T3R/p1741245481738509

Related GDR PR will be pushed soon.
For https://github.com/Kong/go-database-reconciler/issues/166


Prashansa-K:
There are some unit tests that are added for Partials Service. Partials are currently supported only in a few docker images I believe... will fix the test condition, once this is confirmed.

kong/utils.go
```diff
@@ -1102,3 +1234,26 @@ func ClearUnmatchingDeprecations(newPlugin *Plugin, oldPlugin *Plugin, schema ma
...
 }
+
+// FillPartialDefaults ingests partial's defaults from its schema.
+// Takes in a partial struct and mutates it in place.
+func FillPartialDefaults(partial *Partial, schema Schema) error {
+	jsonb, err := json.Marshal(&schema)
+	if err != nil {
+		return err
+	}
+	gjsonSchema := gjson.ParseBytes((jsonb))
```
pmalek, line 1245:
```suggestion
	gjsonSchema := gjson.ParseBytes(jsonb)
```

kong/utils.go
```diff
@@ -797,32 +799,162 @@ func fillConfigRecordDefaultsAutoFields(plugin *Plugin, schema map[string]interf
...
 func FillPluginsDefaults(plugin *Plugin, schema Schema) error {
-	return fillConfigRecordDefaultsAutoFields(plugin, schema, FillRecordOptions{
+	return fillConfigRecordDefaultsAutoFields(plugin, schema, nil, FillRecordOptions{
+		FillDefaults: true,
+		FillAuto:     true,
+	})
+}
+
+// FillPluginsDefaultsWithPartials ingests plugin's defaults from its schema.
+// Takes in a plugin struct and mutate it in place.
```
pmalek, line 939:
```suggestion
// Takes in a plugin struct and mutates it in place.
```

kong/partial_service_test.go
```diff
@@ -0,0 +1,455 @@
...
+	// 			assert.NoError(client.Partials.Delete(defaultCtx, createdPartial.ID))
+	// 		}
+	// 	})
+
+	// 	fetchedPartial, err := client.Partials.Get(defaultCtx, String("my-demo-partial"))
+	// 	assert.NoError(err)
+	// 	assert.NotNil(fetchedPartial)
+	// 	assert.Equal("my-demo-partial", *fetchedPartial.Name)
+	// 	assert.Equal("redis-ee", *fetchedPartial.Type)
+	// })
```
pmalek, line 148: Why is this commented out?

kong/partial_service_test.go
```diff
@@ -0,0 +1,455 @@
...
+			Name: String("my-test-partial"),
+			Type: String("redis-ee"),
+		})
+		require.NoError(err)
+
+		t.Cleanup(func() {
+			if createdPartial != nil {
+				assert.NoError(client.Partials.Delete(defaultCtx, createdPartial.ID))
+			}
+		})
```
pmalek, line 121:
```suggestion
		require.NoError(err)
		require.NotNil(createdPartial)
		t.Cleanup(func() {
			assert.NoError(client.Partials.Delete(defaultCtx, createdPartial.ID))
		})
```

pmalek: And same elsewhere

kong/utils.go
```diff
@@ -797,32 +799,162 @@ func fillConfigRecordDefaultsAutoFields(plugin *Plugin, schema map[string]interf
...
 // same as FillPluginsDefaults but allows configuring whether to fill defaults and auto fields.
 func FillPluginsDefaultsWithOpts(plugin *Plugin, schema map[string]interface{}, opts FillRecordOptions) error {
-	return fillConfigRecordDefaultsAutoFields(plugin, schema, opts)
+	return fillConfigRecordDefaultsAutoFields(plugin, schema, nil, opts)
+}
+
+// FillPluginsDefaultsWithPartials ingests plugin's defaults from its schema.
+// Takes in a plugin struct and mutate it in place.
+// It also fills in partial config
+func FillPluginWithPartials(plugin *Plugin, pluginSchema map[string]interface{}, partials []*Partial) error {
```
pmalek, line 956: Could we add simple UTs for `func`s added in this PR?

At the very least I think we should test the export functions like `FillPluginWithPartials` and `FillPluginsDefaultsWithPartials`

kong/partial_service_test.go
Prashansa-K: Ohh I left it by mistake. I will remove it.
Earlier the APIs were supporting GET, DELETE by name as well. Later, the gateway team only allowed all operations by ID to keep things in sync with Konnect, I suppose. So, this test was added earlier but I later forgot to remove it.

kong/utils.go
```diff
@@ -797,32 +799,168 @@ func fillConfigRecordDefaultsAutoFields(plugin *Plugin, schema map[string]interf
...
+	if plugin.Config == nil {
+		plugin.Config = make(Configuration)
+	}
+
+	if len(partials) > 0 {
+		err = getDefaultPartialPath(plugin.Partials, gjsonSchema, partials)
+		if err != nil {
+			return err
+		}
+		plugin.Config, err = fillPartialConfigsInPlugin(plugin, partials)
```
pmalek, line 926: Nit: Since this func is only called where `plugin.Config` is assigned to the return value, why don't we just perform the assignment in `fillPartialConfigsInPlugin`?

Specifically, what I'm suggesting is to drop the first return value, make the assignment inside the `fillPartialConfigsInPlugin` and potentially do away with the deep copy (I _think_ it won't be necessary at that point).

WDYT?

pmalek (approved):

Issue #519, opened 2025-03-17
Use API endpoint with currently supported Gateway versions in CI


pmalek:
### Problem statement

https://github.com/Kong/go-kong/blob/a905cd98b6058742290fad6aa542916ae3334a8b/.github/workflows/integration-test.yaml#L35-L46 defines a list of Gateway versions against which we test.

This is cumbersome to maintain.

https://github.com/Kong/docs.konghq.com/pull/8561 introduced an API (https://docs.konghq.com/_api/gateway-versions.json) which we can utilize to get an up to date list of currently supported Gateway versions.

This issue tracks adding usage of this API in CI

### Acceptance criteria

- [ ] A workflow which runs on schedule exists, which checks https://docs.konghq.com/_api/gateway-versions.json and creates PRs to update a file in this repo. This file is the source of truth for tested Gateway versions.

mheap:
You'll need to do some parsing of old versions yourself with jq:

```bash
jq '[.[] | select(.endOfLifeDate > (now | strftime("%Y-%m-%d")))]'
```

PR #522, opened 2025-03-19
ci: dynamic kong version matrix

pmalek:
This introduces a dynamic test matrix for Kong versions.

Ideally this would be done with a workflow on schedule so that no unexpected failures happen in unrelated CI but 1 thing at a time.

Part of #519

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,16 @@ golangci-lint: mise yq ## Download golangci-lint locally if necessary.
 	@$(MISE) install -q golangci-lint@$(GOLANGCI_LINT_VERSION)
 
 # ------------------------------------------------------------------------------
+# CI
+# ------------------------------------------------------------------------------
+
+
+.PHONY: kong.supported-versions
+kong.supported-versions:
+	@curl -s https://docs.konghq.com/_api/gateway-versions.json | \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -30,6 +30,16 @@ golangci-lint: mise yq ## Download golangci-lint locally if necessary.
...
 
+# ------------------------------------------------------------------------------
+# CI
+# ------------------------------------------------------------------------------
+
+
+.PHONY: kong.supported-versions
+kong.supported-versions:
+	@curl -s https://docs.konghq.com/_api/gateway-versions.json | \
+		jq '[.[] | select(.endOfLifeDate > (now | strftime("%Y-%m-%d")))] | [.[].tag]'
```
czeslavo, line 41: An unreleased version could be returned from this endpoint. E.g. now `3.10` is not returned from this `jq` query because it has no `endOfLifeDate`, but I'm unsure if this will always be true. Maybe we could explicitly filter out entries with `label: unreleased`?

pmalek: Good catch. 36fc97cd83248274c5ef8f7f8d4617e9314d5078 should handle that.

czeslavo (approved):

PR #526, opened 2025-03-24
Add GetByID interface for credentials - such as ACL, basic auth etc

harshadixit12:
These changes are being made to support the fix for https://github.com/Kong/deck/issues/1543.

Changes made:
- GetByID function added to the credential service, and each credential
- Tests

https://kongstrong.slack.com/archives/CQK8J4VN3/p1741703868872269

Skipped mtls auth - gateway does not seem to support Get By ID endpoint.

### DIFF
--- a/kong/acl_service_test.go
+++ b/kong/acl_service_test.go
@@ -137,7 +137,7 @@ func TestACLGroupGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
-func TestACLGroupGetById(T *testing.T) {
+func TestACLGroupGetByID(T *testing.T) {
 	RunWhenDBMode(T, "postgres")
 
 	assert := assert.New(T)
@@ -166,15 +166,15 @@ func TestACLGroupGetById(T *testing.T) {
 	require.NoError(err)
 	assert.NotNil(createdACL)
 
-	aclGroup, err := client.ACLs.GetById(defaultCtx, acl.ID)
+	aclGroup, err := client.ACLs.GetByID(defaultCtx, acl.ID)
 	require.NoError(err)
 	assert.Equal("my-group", *aclGroup.Group)
 
-	aclGroup, err = client.ACLs.GetById(defaultCtx, String("does-not-exist"))
+	aclGroup, err = client.ACLs.GetByID(defaultCtx, String("does-not-exist"))
 	assert.Nil(aclGroup)
 	require.Error(err)
 
-	aclGroup, err = client.ACLs.GetById(defaultCtx, String(""))
+	aclGroup, err = client.ACLs.GetByID(defaultCtx, String(""))
 	assert.Nil(aclGroup)
 	require.Error(err)
 
--- a/kong/acl_service_test.go
+++ b/kong/acl_service_test.go
@@ -137,6 +137,50 @@ func TestACLGroupGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
+func TestACLGroupGetById(T *testing.T) {
+	RunWhenDBMode(T, "postgres")
+
+	assert := assert.New(T)
+	require := require.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	require.NoError(err)
+	assert.NotNil(client)
+
+	uuid := uuid.NewString()
+	acl := &ACLGroup{
+		ID:    String(uuid),
+		Group: String("my-group"),
+	}
+
+	// consumer for the ACLGroup
+	consumer := &Consumer{
+		Username: String("foo"),
+	}
+
+	consumer, err = client.Consumers.Create(defaultCtx, consumer)
+	require.NoError(err)
+	require.NotNil(consumer)
+
+	createdACL, err := client.ACLs.Create(defaultCtx, consumer.ID, acl)
+	require.NoError(err)
+	assert.NotNil(createdACL)
+
+	aclGroup, err := client.ACLs.GetById(defaultCtx, acl.ID)
+	require.NoError(err)
+	assert.Equal("my-group", *aclGroup.Group)
+
+	aclGroup, err = client.ACLs.GetById(defaultCtx, String("does-not-exist"))
+	assert.Nil(aclGroup)
+	require.Error(err)
+
+	aclGroup, err = client.ACLs.GetById(defaultCtx, String(""))
+	assert.Nil(aclGroup)
+	require.Error(err)
+
+	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
+}
+
 func TestACLGroupUpdate(T *testing.T) {
 	RunWhenDBMode(T, "postgres")
 
--- a/kong/acl_service_test.go
+++ b/kong/acl_service_test.go
@@ -137,14 +137,12 @@ func TestACLGroupGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
-func TestACLGroupGetByID(T *testing.T) {
-	RunWhenDBMode(T, "postgres")
-
-	require := require.New(T)
+func TestACLGroupGetByID(t *testing.T) {
+	RunWhenDBMode(t, "postgres")
 
 	client, err := NewTestClient(nil, nil)
-	require.NoError(err)
-	require.NotNil(client)
+	require.NoError(t, err)
+	require.NotNil(t, client)
 
 	uuid := uuid.NewString()
 	acl := &ACLGroup{
@@ -158,34 +156,34 @@ func TestACLGroupGetByID(T *testing.T) {
 	}
 
 	consumer, err = client.Consumers.Create(defaultCtx, consumer)
-	require.NoError(err)
-	require.NotNil(consumer)
+	require.NoError(t, err)
+	require.NotNil(t, consumer)
 
 	createdACL, err := client.ACLs.Create(defaultCtx, consumer.ID, acl)
-	require.NoError(err)
-	require.NotNil(createdACL)
+	require.NoError(t, err)
+	require.NotNil(t, createdACL)
 
-	T.Cleanup(func() {
-		require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
+	t.Cleanup(func() {
+		require.NoError(t, client.Consumers.Delete(defaultCtx, consumer.ID))
 	})
 
-	T.Run("successful ACL retrieval by ID", func(_ *testing.T) {
+	t.Run("successful ACL retrieval by ID", func(_ *testing.T) {
 		aclGroup, err := client.ACLs.GetByID(defaultCtx, acl.ID)
-		require.NoError(err)
-		require.NotNil(aclGroup)
-		require.Equal("my-group", *aclGroup.Group)
+		require.NoError(t, err)
+		require.NotNil(t, aclGroup)
+		require.Equal(t, "my-group", *aclGroup.Group)
 	})
 
-	T.Run("unsuccessful ACL retrieval using invalid ID", func(_ *testing.T) {
+	t.Run("unsuccessful ACL retrieval using invalid ID", func(_ *testing.T) {
 		aclGroup, err := client.ACLs.GetByID(defaultCtx, String("does-not-exist"))
-		require.Nil(aclGroup)
-		require.Error(err)
+		require.Nil(t, aclGroup)
+		require.Error(t, err)
 	})
 
-	T.Run("unsuccessful ACL retrieval using empty string", func(_ *testing.T) {
+	t.Run("unsuccessful ACL retrieval using empty string", func(_ *testing.T) {
 		aclGroup, err := client.ACLs.GetByID(defaultCtx, String(""))
-		require.Nil(aclGroup)
-		require.Error(err)
+		require.Nil(t, aclGroup)
+		require.Error(t, err)
 	})
 }
 
--- a/kong/acl_service_test.go
+++ b/kong/acl_service_test.go
@@ -167,20 +167,20 @@ func TestACLGroupGetByID(t *testing.T) {
 		require.NoError(t, client.Consumers.Delete(defaultCtx, consumer.ID))
 	})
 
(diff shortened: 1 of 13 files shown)


kong/acl_service_test.go
```diff
@@ -137,6 +137,50 @@ func TestACLGroupGet(T *testing.T) {
...
 
+func TestACLGroupGetByID(T *testing.T) {
+	RunWhenDBMode(T, "postgres")
+
+	assert := assert.New(T)
+	require := require.New(T)
+
+	client, err := NewTestClient(nil, nil)
+	require.NoError(err)
+	assert.NotNil(client)
```
Prashansa-K, line 148:
```suggestion
	require.NotNil(client)
```
Using `require` here would be better to avoid potential panics. Let's make this change in other cred tests too (the ones you are adding in this PR).

kong/acl_service_test.go
```diff
@@ -137,6 +137,50 @@ func TestACLGroupGet(T *testing.T) {
...
+
+	aclGroup, err = client.ACLs.GetByID(defaultCtx, String("does-not-exist"))
+	assert.Nil(aclGroup)
+	require.Error(err)
+
+	aclGroup, err = client.ACLs.GetByID(defaultCtx, String(""))
+	assert.Nil(aclGroup)
+	require.Error(err)
+
+	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
```
Prashansa-K, line 181: Use `t.Cleanup()` for such operations. That makes the intent clear.
We should also cleanup the other entity that you created here - the cred.
Refer: https://github.com/Kong/go-kong/blob/main/kong/vault_service_test.go#L30

kong/acl_service_test.go
```diff
@@ -137,6 +137,50 @@ func TestACLGroupGet(T *testing.T) {
...
+	consumer, err = client.Consumers.Create(defaultCtx, consumer)
+	require.NoError(err)
+	require.NotNil(consumer)
+
+	createdACL, err := client.ACLs.Create(defaultCtx, consumer.ID, acl)
+	require.NoError(err)
+	assert.NotNil(createdACL)
+
+	aclGroup, err := client.ACLs.GetByID(defaultCtx, acl.ID)
+	require.NoError(err)
```
Prashansa-K, line 170:
```suggestion
	require.NoError(err)
	require.NotNil(aclGroup)
```

kong/credentials_service.go
```diff
@@ -109,6 +121,32 @@ func (s *credentialService) Get(ctx context.Context, credType string,
 	return cred, nil
 }
 
+// GetByID fetches a credential of credType with ID from Kong.
+func (s *credentialService) GetByID(ctx context.Context, credType string, credIdentifier *string,
+) (json.RawMessage, error) {
+	if isEmptyString(credIdentifier) {
```
Prashansa-K, line 127: Shouldn't we add a similar check for `credType`?

kong/acl_service_test.go
Prashansa-K: This should be changed in all places.

kong/acl_service_test.go
Prashansa-K: Add similar change in other tests as well.

kong/acl_service_test.go
```diff
@@ -137,6 +137,50 @@ func TestACLGroupGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
+func TestACLGroupGetByID(T *testing.T) {
```
Prashansa-K, line 140: Since you are testing multiple use-cases in one test function, I would suggest to make sub-tests via `t.Run()`
It is better for code readability and also tells us which test is failing precisely.
Can refer this for the same: https://github.com/Kong/go-kong/blob/main/kong/partial_service_test.go

Prashansa-K (changes requested):

kong/credentials_service.go
harshadixit12: We are checking for a valid credType on line 132 - https://github.com/Kong/go-kong/pull/526/files#diff-c7a9e15599f12d918c2a99b837688ce87f76a4a996e369301e424ffa473b24ccR132
Since empty string is a valid key in a map - IMO additional check is not needed

Prashansa-K (approved):

kong/acl_service_test.go
```diff
@@ -137,6 +137,58 @@ func TestACLGroupGet(T *testing.T) {
...
```
czeslavo, line 140: There's no reason to name a local variable with a capital letter - let's just use `t` as that's the common pattern.
```suggestion
func TestACLGroupGetByID(t *testing.T) {
```

kong/basic_auth_service.go
```diff
@@ -12,6 +12,8 @@ type AbstractBasicAuthService interface {
 	Create(ctx context.Context, consumerUsernameOrID *string, basicAuth *BasicAuth) (*BasicAuth, error)
 	// Get fetches a basic-auth credential from Kong.
 	Get(ctx context.Context, consumerUsernameOrID, usernameOrID *string) (*BasicAuth, error)
+	// GetByID fetches a basic-auth credential from Kong using ID.
+	GetByID(ctx context.Context, ID *string) (*BasicAuth, error)
```
czeslavo, line 16: Ditto
```suggestion
	GetByID(ctx context.Context, id *string) (*BasicAuth, error)
```

kong/basic_auth_service.go
```diff
@@ -69,6 +71,25 @@ func (s *BasicAuthService) Get(ctx context.Context,
 	return &basicAuth, nil
 }
 
+// GetByID fetches a basic-auth credential from Kong using ID.
+func (s *BasicAuthService) GetByID(ctx context.Context,
+	ID *string,
```
czeslavo, line 76: Ditto.
```suggestion
	id *string,
```

kong/basic_auth_service_test.go
```diff
@@ -155,6 +155,60 @@ func TestBasicAuthGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
+func TestBasicAuthGetByID(T *testing.T) {
```
czeslavo, line 158: Ditto.
```suggestion
func TestBasicAuthGetByID(t *testing.T) {
```

kong/hmac_auth_service.go
```diff
@@ -11,6 +11,8 @@ type AbstractHMACAuthService interface {
 	Create(ctx context.Context, consumerUsernameOrID *string, hmacAuth *HMACAuth) (*HMACAuth, error)
 	// Get fetches a hmac-auth credential from Kong.
 	Get(ctx context.Context, consumerUsernameOrID, usernameOrID *string) (*HMACAuth, error)
+	// GetByID fetches a hmac-auth credential from Kong using hmac-auth ID.
+	GetByID(ctx context.Context, ID *string) (*HMACAuth, error)
```
czeslavo, line 15: Ditto.
```suggestion
	GetByID(ctx context.Context, id *string) (*HMACAuth, error)
```

kong/hmac_auth_service_test.go
```diff
@@ -144,6 +144,59 @@ func TestHMACAuthGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
+func TestHMACAuthGetByID(T *testing.T) {
```
czeslavo, line 147: Ditto.
```suggestion
func TestHMACAuthGetByID(t *testing.T) {
```

kong/jwt_auth_service.go
```diff
@@ -11,6 +11,8 @@ type AbstractJWTAuthService interface {
 	Create(ctx context.Context, consumerUsernameOrID *string, jwtAuth *JWTAuth) (*JWTAuth, error)
 	// Get fetches a JWT credential from Kong.
 	Get(ctx context.Context, consumerUsernameOrID, keyOrID *string) (*JWTAuth, error)
+	// GetByID fetches a JWT credential from Kong using ID.
+	GetByID(ctx context.Context, ID *string) (*JWTAuth, error)
```
czeslavo, line 15: Ditto.
```suggestion
	GetByID(ctx context.Context, id *string) (*JWTAuth, error)
```

kong/jwt_auth_service.go
```diff
@@ -68,6 +70,25 @@ func (s *JWTAuthService) Get(ctx context.Context,
 	return &jwtAuth, nil
 }
 
+// GetByID fetches a JWT credential from Kong using ID.
+func (s *JWTAuthService) GetByID(ctx context.Context,
+	ID *string,
```
czeslavo, line 75: Ditto.

```suggestion
	id *string,
```

kong/jwt_auth_service_test.go
```diff
@@ -143,6 +143,58 @@ func TestJWTGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
+func TestJWTGetByID(T *testing.T) {
```
czeslavo, line 146:
```suggestion
func TestJWTGetByID(t *testing.T) {
```

kong/key_auth_service.go
```diff
@@ -11,6 +11,8 @@ type AbstractKeyAuthService interface {
 	Create(ctx context.Context, consumerUsernameOrID *string, keyAuth *KeyAuth) (*KeyAuth, error)
 	// Get fetches a key-auth credential from Kong.
 	Get(ctx context.Context, consumerUsernameOrID, keyOrID *string) (*KeyAuth, error)
+	// GetByID fetches a key-auth credential from Kong using ID
+	GetByID(ctx context.Context, ID *string) (*KeyAuth, error)
```
czeslavo, line 15:
```suggestion
	GetByID(ctx context.Context, id *string) (*KeyAuth, error)
```

kong/key_auth_service.go
```diff
@@ -67,6 +69,24 @@ func (s *KeyAuthService) Get(ctx context.Context,
 	return &keyAuth, nil
 }
 
+// GetByID fetches a key-auth credential from Kong using ID.
+func (s *KeyAuthService) GetByID(ctx context.Context, ID *string,
```
czeslavo, line 73:
```suggestion
func (s *KeyAuthService) GetByID(ctx context.Context, id *string,
```

kong/key_auth_service_test.go
```diff
@@ -141,6 +141,59 @@ func TestKeyAuthGet(T *testing.T) {
 	require.NoError(client.Consumers.Delete(defaultCtx, consumer.ID))
 }
 
+func TestKeyAuthGetByID(T *testing.T) {
```
czeslavo, line 144:
```suggestion
func TestKeyAuthGetByID(t *testing.T) {
```

kong/oauth2_auth_service.go
```diff
@@ -11,6 +11,8 @@ type AbstractOauth2Service interface {
 	Create(ctx context.Context, consumerUsernameOrID *string, oauth2Cred *Oauth2Credential) (*Oauth2Credential, error)
 	// Get fetches an oauth2 credential from Kong.
 	Get(ctx context.Context, consumerUsernameOrID, clientIDorID *string) (*Oauth2Credential, error)
+	// GetByID fetches an oauth2 credential from Kong using ID.
+	GetByID(ctx context.Context, ID *string) (*Oauth2Credential, error)
```
czeslavo, line 15:
```suggestion
	GetByID(ctx context.Context, id *string) (*Oauth2Credential, error)
```

kong/oauth2_auth_service.go
```diff
@@ -69,6 +71,25 @@ func (s *Oauth2Service) Get(ctx context.Context,
 	return &oauth2Cred, nil
 }
 
+// GetByID fetches an oauth2 credential from Kong using ID.
+func (s *Oauth2Service) GetByID(ctx context.Context,
+	ID *string,
```
czeslavo, line 76:
```suggestion
	id *string,
```

kong/acl_service_test.go
```diff
@@ -137,6 +137,58 @@ func TestACLGroupGet(T *testing.T) {
...
+	RunWhenDBMode(T, "postgres")
+
+	require := require.New(T)
```
czeslavo, line 143: I suggest omitting this `require` instantiation here and instead relying directly on the `require` package functions. 

Currently, in every sub-case you're running with `t.Run(),` you're relying on an instance of `require` that was created with a `testing.T` from the top-level case. This makes you not use the `testing.T` that has a context of the sub-case and allows better reporting if an assertion fails (with the top-level `testing.T` you'll get a failure in the top-level case instead of a sub-case).

It's correct to run sub-cases like so:
```go
	t.Run("successful ACL retrieval by ID", func(t *testing.T) {
		aclGroup, err := client.ACLs.GetByID(defaultCtx, acl.ID)
		require.NoError(t, err)
		require.NotNil(t, aclGroup)
		require.Equal(t, "my-group", *aclGroup.Group)
	})
```

czeslavo (commented):
Leaving comments mostly on code cosmetics. From the API standpoint it looks good to me.

kong/acl_service_test.go
harshadixit12: Done, the existing tests had the variable with capital letter, so I had followed the same initially.

kong/basic_auth_service_test.go
```diff
@@ -155,6 +155,58 @@ func TestBasicAuthGet(T *testing.T) {
...
+	createdBasicAuth, err := client.BasicAuths.Create(defaultCtx,
+		consumer.ID, basicAuth)
+	require.NoError(t, err)
+	require.NotNil(t, createdBasicAuth)
+
+	t.Cleanup(func() {
+		require.NoError(t, client.Consumers.Delete(defaultCtx, consumer.ID))
+	})
+
+	t.Run("successful basic-auth retrieval by ID", func(_ *testing.T) {
```
czeslavo, line 190: We should use `testing.T` of this sub-test case instead of the top-level one. Same applies to every subtest.
```suggestion
	t.Run("successful basic-auth retrieval by ID", func(t *testing.T) {
```

czeslavo (approved):

PR #533, opened 2025-04-07
feat: use pagination to load all pages of rbac endpoint permissions for a role

harshadixit12:
Changes:
feat: use pagination to load all pages of rbac endpoint permissions for a role


kong/endpoint_permission_service_test.go
```diff
@@ -87,3 +88,56 @@ func TestRBACEndpointPermissionservice(T *testing.T) {
...
+	workspaceClient, err := NewTestClient(
+		String(defaultBaseURL+"/rbac-endpoint-permission-list-endpoint-test-workspace"), nil)
+	require.NoError(t, err)
+
+	role, err := workspaceClient.RBACRoles.Create(defaultCtx, &RBACRole{Name: String("roleA")})
+	require.NoError(t, err)
+	assert.NotNil(t, role)
+
+	t.Run("load all pages", func(t *testing.T) {
+		for i := 0; i < pageSize+1; i++ {
```
harshadixit12, line 117: Creating 1001 endpoint permissions seems a little slow - I'm open to suggestions if there is a faster way of doing it.

kong/endpoint_permission_service_test.go
```diff
@@ -87,3 +88,56 @@ func TestRBACEndpointPermissionservice(T *testing.T) {
...
+		}
+
+		endpointPermissionsFromKong, err := workspaceClient.RBACEndpointPermissions.ListAllForRole(defaultCtx, role.ID)
+		require.NoError(t, err)
+		assert.NotNil(t, endpointPermissionsFromKong)
+		assert.Len(t, endpointPermissionsFromKong, pageSize+1)
+
+		err = workspaceClient.RBACRoles.Delete(defaultCtx, role.ID)
+		require.NoError(t, err)
+		err = client.Workspaces.Delete(defaultCtx, createdWorkspace.ID)
```
harshadixit12, line 140: Not deleting the endpoint permissions one by one since they get deleted anyway on workspace delete - please correct me if wrong.

harshadixit12:
CI is failing on nightly builds, I'll find out why

kong/endpoint_permission_service_test.go
czeslavo: It should be enough if you create only 2 endpoints as the `pageSize` used in the client is set on tests startup to 1: https://github.com/Kong/go-kong/blob/f07287df3813c28cb3b97989e3a5c7333abb6203/kong/client_test.go#L169

kong/endpoint_permission_service_test.go
```diff
@@ -87,3 +88,56 @@ func TestRBACEndpointPermissionservice(T *testing.T) {
...
 }
+
+func TestRBACEndpointPermissionList(t *testing.T) {
+	RunWhenEnterprise(t, ">=0.33.0", RequiredFeatures{RBAC: true})
+
+	client, err := NewTestClient(nil, nil)
+	require.NoError(t, err)
+	assert.NotNil(t, client)
+
+	pageSize := 1000
```
czeslavo, line 99: Let's rely on the package-level `pageSize` variable instead of shadowing it here.

czeslavo (commented):
LGTM, just two nits regarding `pageSize` in tests.

czeslavo (approved):

PR #542, opened 2025-05-09
feat: Implement FillID method for plugins

randmonkey:
Implement `FillID` method for Plugin for https://github.com/Kong/kubernetes-ingress-controller/issues/7394. 
Because GDR regards plugin with the same name, service, route, consumer, and consumer group as the same plugin, the method for generating ID takes them into account.

### DIFF
--- a/kong/ids.go
+++ b/kong/ids.go
@@ -128,6 +128,49 @@ func (v *Vault) FillID(workspace string) error {
 	return nil
 }
 
+// FillID fills the ID of an entity. It is a no-op if the entity already has an ID.
+// ID is generated in a deterministic way using UUIDv5.
+// The UUIDv5 namespace being used for generation is separate from other namespaces used for generating IDs for other types.
+// The name used to generate the ID for plugins is the combination of:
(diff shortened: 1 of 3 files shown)


kong/ids.go
```diff
@@ -128,6 +128,49 @@ func (v *Vault) FillID(workspace string) error {
...
+	}
+	if p.Route != nil {
+		toHash = toHash + ":route/" + p.Route.FriendlyName()
+	}
+	if p.Consumer != nil {
+		toHash = toHash + ":consumer/" + p.Consumer.FriendlyName()
+	}
+	if p.ConsumerGroup != nil {
+		toHash = toHash + ":consumer_group/" + p.ConsumerGroup.FriendlyName()
+	}
```
pmalek, line 163: Nit: to make this ever so slightly more efficient we could use `strings.Builder` to reduce the number of string copies: https://pkg.go.dev/strings#Builder. WDYT?

randmonkey: Done.

pmalek (approved):

PR #543, opened 2025-05-09
Set default value of `config.throttling` field of RLA plugin in integration tests when Kong version >= 3.11

randmonkey:
Kong Enterprise added `throttling` field to the `rate-limiting-advanced` plugin and assigned default values for it: https://github.com/Kong/kong-ee/pull/12243

The PR sets the default value of `config.throttling` field of RLA plugin if the Kong version supports (>= 3.11).

### DIFF
--- a/kong/test_utils.go
+++ b/kong/test_utils.go
@@ -45,7 +45,7 @@ func RunWhenKong(t *testing.T, versionRange string) {
 		t.Error(err)
 	}
 	if !r(currentVersion) {
-		t.Skipf("kong version %s not in range %s", currentVersion.String(), versionRange)
+		t.Skipf("kong version %s not in range %s", currentVersion, versionRange)
 	}
 }
 
--- a/kong/test_utils.go
+++ b/kong/test_utils.go
@@ -12,11 +12,8 @@ type RequiredFeatures struct {
 	RBAC   bool
 }
 
-// RunWhenKong skips the current test if the version of Kong doesn't
-// fall in the versionRange.
-// This helper function can be used in tests to write version specific
-// tests for Kong.
-func RunWhenKong(t *testing.T, versionRange string) {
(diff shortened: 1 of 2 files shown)


kong/utils_test.go
```diff
@@ -4209,6 +4209,12 @@ func Test_FillPluginsDefaultsWithPartials(t *testing.T) {
 			errString: "no 'config' field found in schema",
 		},
 	}
+	// Kong Enterprise added `throttling` field and assigned default values since 3.11.
+	// We ignore the field here to be compatible with 3.10.
+	// TODO: set default value in the expected plugin configuration when Kong version >= 3.11?
```
pmalek, line 4214: I believe we should set this default in the expected plugin configuration for 3.11+ and not ignore it in the test assertion below. WDYT?

kong/test_utils.go
```diff
@@ -32,12 +29,23 @@ func RunWhenKong(t *testing.T, versionRange string) {
...
+	t.Helper()
+
+	currentVersion := GetVersionForTesting(t)
 	r, err := NewRange(versionRange)
 	if err != nil {
 		t.Error(err)
 	}
 	if !r(currentVersion) {
-		t.Skipf("kong version %s not in range %s", version, versionRange)
+		t.Skipf("kong version %s not in range %s", currentVersion.String(), versionRange)
```
pmalek, line 48: Nit: 

No need to call `.String()` when using the `%s` formatting string.

```suggestion
		t.Skipf("kong version %s not in range %s", currentVersion, versionRange)
```

pmalek (approved):
1 minor nit, otherwise this looks good.

It would be nice to link to the defaults use here (in docs or elsewhere) but since 3.11 hasn't been released yet they don't exist so I think we can merge as is.

Issue #554, opened 2025-06-18
Add support for Kong "/upstreams/{upstream_id_or_name}/targets/{target_id_or_target}" PATCH and PUT API


DWonMtl:
# Summary 

The current Kong Admin API for Targets supports the following HTTP methods:
- [GET](https://docs.konghq.com/gateway/api/admin-ee/3.10.0.x/#/operations/fetch-target-with-upstream)
- [DELETE](https://docs.konghq.com/gateway/api/admin-ee/3.10.0.x/#/operations/delete-target-with-upstream)
- [POST](https://docs.konghq.com/gateway/api/admin-ee/3.10.0.x/#/operations/create-target-with-upstream)
- [PATCH](https://docs.konghq.com/gateway/api/admin-ee/3.10.0.x/#/operations/update-target-with-upstream)
- [PUT](https://docs.konghq.com/gateway/api/admin-ee/3.10.0.x/#/operations/upsert-target-with-upstream)

However, the current go-kong client implementation only supports:
- [Create](https://github.com/Kong/go-kong/blob/v0.66.1/kong/target_service.go#L12) (related to POST)
- [Delete](https://github.com/Kong/go-kong/blob/v0.66.1/kong/target_service.go#L14) (related to DELETE)
- List 
- ListAll
- MarkHealthy
- MarkUnhealthy

# Proposal

I propose adding an `Update` method to the `TargetService` using the `PATCH` operation, enabling partial updates to existing targets:

```golang
	// Update updates a Target in Kong
	Update(ctx context.Context, upstreamNameOrID *string, targetOrID *string, target *Target) (*Target, error)
```

This would allow modifying fields such as weight and tags without requiring a full delete-and-recreate cycle.

# Additional Notes 

There is a [comment in the code suggesting the potential use of PUT in the Create method](https://github.com/Kong/go-kong/blob/v0.66.1/kong/target_service.go#L45). This was previously tracked by a [TODO](https://github.com/Kong/go-kong/commit/8cb22d2cfd5feef33f22ff0959374a69bfcd5048#diff-efb65d11f1a5bf899e1c94b15fac43e2b84e91a5ea39eb52e9b9bed4720088a7R23), but it has not yet been implemented.

DWonMtl:
Hi @pmalek — I noticed you've been the primary contributor since 2022. I'd be happy to help with this issue if you're open to collaboration.

PR #555, opened 2025-06-19
Feat/554 add support for kong target patch and put api

DWonMtl:
Following this issue : https://github.com/Kong/go-kong/issues/554

This PR impelment a new Kong Target `Update` function to the `TargetService` 

```golang
	// Update updates a Target in Kong
	Update(ctx context.Context, upstreamNameOrID *string, targetOrID *string, target *Target) (*Target, error)
```

This also implement the following comments https://github.com/Kong/go-kong/blob/v0.66.1/kong/target_service.go#L45 desire to add `PUT` support in the `TargetService.Create` function.

To avoid breaking change the comment is not applied as is since, right now, If the `target.ID` is configured but does not exist, the Target will be created. Changing it to `PUT` as stated would return an error.

Therefore, the implementation is a little different. Where, if someone specify a `target.ID`, `PUT` will be tried, and if the Target does not exist, then it will be created, otherwise it will simply be updated.

Ran tests and linters locally


kong/target_service.go
```diff
@@ -41,11 +43,30 @@ func (s *TargetService) Create(ctx context.Context,
 		return nil, fmt.Errorf("upstreamNameOrID can not be nil")
 	}
 	queryPath := "/upstreams/" + *upstreamNameOrID + "/targets"
+
+	// If the target has an ID, we will try to update it.
+	if target.ID != nil {
+		method := "PUT"
+		req, err := s.client.NewRequest(method, queryPath+"/"+*target.ID, nil, target)
```
pmalek, line 50:
```suggestion
		req, err := s.client.NewRequest("PUT", queryPath+"/"+*target.ID, nil, target)
```

kong/target_service.go
```diff
@@ -41,11 +43,30 @@ func (s *TargetService) Create(ctx context.Context,
...
+		}
+	}
+
+	// If the target does not have an ID OR does not exist, we will create it.
 	method := "POST"
-	// if target.ID != nil {
-	// 	queryPath = queryPath + "/" + *target.ID
-	// 	method = "PUT"
-	// }
 	req, err := s.client.NewRequest(method, queryPath, nil, target)
```
pmalek, line 70:
```suggestion
	req, err := s.client.NewRequest("POST", queryPath, nil, target)
```

kong/target_service.go
```diff
@@ -41,11 +43,30 @@ func (s *TargetService) Create(ctx context.Context,
...
+		if !IsNotFoundErr(err) {
+			if err != nil {
+				return nil, err
+			}
+			return &createdTarget, nil
+		}
+	}
+
+	// If the target does not have an ID OR does not exist, we will create it.
 	method := "POST"
```
pmalek, line 69:
```suggestion
```

kong/target_service.go
```diff
@@ -22,6 +22,8 @@ type AbstractTargetService interface {
 	// MarkUnhealthy marks target belonging to upstreamNameOrID as unhealthy in
 	// Kong's load balancer.
 	MarkUnhealthy(ctx context.Context, upstreamNameOrID *string, target *Target) error
+	// Update updates a Target in Kong under upstreamID.
```
pmalek, line 25: Let's make sure to add a note that `PATCH` will be used for the update operation.

pmalek (commented):
Some minor nits, otherwise this looks good.

Please make sure to add a changelog entry for this.

kong/target_service.go
DWonMtl: Done in commit https://github.com/Kong/go-kong/commit/6103a8829bd51bc9f0c81d9c106a582b5a38d6d6

DWonMtl:
> Some minor nits, otherwise this looks good.
> 
> Please make sure to add a changelog entry for this.

Thanks for the review! 

I tried to write a changelog entry. But here I am not super familiar with the project process. Let me know if there are anything that must be changed. (Done in commit https://github.com/Kong/go-kong/commit/6103a8829bd51bc9f0c81d9c106a582b5a38d6d6)

kong/target_service_test.go
```diff
@@ -93,20 +94,37 @@ func TestTargetsUpdate(T *testing.T) {
 	assert.NotNil(createdTarget)
 	assert.Equal(targetID, *createdTarget.ID)
 
+	// determine the target with predefined ID
```
pmalek, line 97:
```suggestion
	// delete the target with predefined ID
```

?

I'm not so found of using comments in tests as they do not surface during tests runs. If we do need something to be added here I'd much rather prefer to see a log which would actually be of value when going through tests output.

DWonMtl: Sorry, I added the comments mostly as a personal reminder when doing the test to "split" it into section. I removed the comment as the code is relatively self-explanatory. 

Done in commit 92b49b0352904798529f112625db85a44425393e

kong/target_service_test.go
```diff
@@ -223,6 +237,53 @@ func TestTargetListEndpoint(T *testing.T) {
...
+
+	targetID := "0fa49cd2-ee93-492a-bedf-b80778d539ae"
+	createdTarget, err := client.Targets.Create(defaultCtx,
+		fixtureUpstream.ID, &Target{
+			ID:     &targetID,
+			Weight: Int(100),
+			Target: String("10.0.0.1:80"),
+		})
+	require.NoError(err)
+	require.NotNil(createdTarget)
```
pmalek, line 267:
```suggestion
	require.NotNil(createdTarget)
	require.NotNil(createdTarget.ID)
	require.NotNil(createdTarget.Weight)
```

kong/target_service_test.go
```diff
@@ -223,6 +237,53 @@ func TestTargetListEndpoint(T *testing.T) {
...
+	assert.Equal(100, *createdTarget.Weight)
+
+	updatedTarget, err := client.Targets.Update(defaultCtx,
+		fixtureUpstream.ID, createdTarget.ID, &Target{
+			ID:     createdTarget.ID,
+			Weight: Int(10000), // Update weight
+			Target: createdTarget.Target,
+		})
+	require.NoError(err)
+	require.NotNil(updatedTarget)
```
pmalek, line 278:
```suggestion
	require.NotNil(updatedTarget)
	require.NotNil(updatedTarget.Target)
	require.NotNil(updatedTarget.Weight)
```

kong/target_service_test.go
```diff
@@ -223,6 +237,53 @@ func TestTargetListEndpoint(T *testing.T) {
...
+	assert.Equal(targetID, *updatedTarget.ID)
+	assert.Equal("10.0.0.1:80", *updatedTarget.Target)
+	assert.Equal(10000, *updatedTarget.Weight)
+
+	err = client.Upstreams.Delete(defaultCtx, fixtureUpstream.ID)
+	require.NoError(err)
```
pmalek, line 284: Please move this above just after this is created.

The reason for this is that when the test fails on any of the `require` clauses above we still want to ensure we call the delete.

kong/target_service_test.go
```diff
@@ -100,13 +100,27 @@ func TestTargetsUpdate(T *testing.T) {
...
 	assert.Equal("10.0.0.2:80", *createdTarget.Target)
 
+	updatedTarget, err := client.Targets.Create(defaultCtx,
+		fixtureUpstream.ID, &Target{
+			ID:     &targetID,
+			Weight: Int(2),
+			Target: String("10.0.0.3:80"),
+		})
+	require.NoError(err)
+	require.NotNil(updatedTarget)
```
pmalek, line 119:
```suggestion
	require.NotNil(updatedTarget)
	require.NotNil(updatedTarget.ID)
	require.NotNil(updatedTarget.Target)
	require.NotNil(updatedTarget.Weight)
```

kong/target_service_test.go
```diff
@@ -100,13 +100,27 @@ func TestTargetsUpdate(T *testing.T) {
...
 	assert.Equal(targetID, *createdTarget.ID)
+	assert.Equal(1, *createdTarget.Weight)
 	assert.Equal("10.0.0.2:80", *createdTarget.Target)
 
+	updatedTarget, err := client.Targets.Create(defaultCtx,
+		fixtureUpstream.ID, &Target{
+			ID:     &targetID,
+			Weight: Int(2),
+			Target: String("10.0.0.3:80"),
+		})
```
pmalek, line 117: Add a delete for this target using `t.Cleanup()`

kong/target_service_test.go
```diff
@@ -100,13 +100,27 @@ func TestTargetsUpdate(T *testing.T) {
...
+	require.NoError(err)
+	require.NotNil(updatedTarget)
+	assert.Equal(targetID, *updatedTarget.ID)
+	assert.Equal(2, *updatedTarget.Weight)
+	assert.Equal("10.0.0.3:80", *updatedTarget.Target)
+
 	err = client.Upstreams.Delete(defaultCtx, fixtureUpstream.ID)
```
pmalek, line 124: While we're at it we might as well move this delete above also using `t.Cleanup()`

kong/target_service_test.go
DWonMtl: Done in commit 288afd55f61de7fe2cb6f6ec0bded00d20544e39

I used a T.Cleanup to be sure the Update is done on an existing Target.

kong/target_service_test.go
DWonMtl: Done

PR #576, opened 2025-09-24
feat: added skip-hash option for basic-auth creation

Prashansa-K:


kong/basic_auth_service.go
```diff
@@ -38,7 +52,39 @@ func (s *BasicAuthService) Create(ctx context.Context,
 	consumerUsernameOrID *string, basicAuth *BasicAuth,
 ) (*BasicAuth, error) {
 	cred, err := s.client.credentials.Create(ctx, "basic-auth",
-		consumerUsernameOrID, basicAuth)
+		consumerUsernameOrID, basicAuth, false)
```
randmonkey, line 55: Could we call `CreateWithOptions` inside `Create` to make code simpler?

kong/credentials_service.go
```diff
@@ -11,15 +11,15 @@ import (
 type abstractCredentialService interface {
 	// Create creates a credential in Kong of type credType.
 	Create(ctx context.Context, credType string, consumerUsernameOrID *string,
-		credential interface{}) (json.RawMessage, error)
-	// Get fetches a credential of credType with consumerIdentifier and credIdentifier from Kong.
+		credential interface{}, skipHash bool) (json.RawMessage, error)
```
randmonkey, line 14: Could we define types such as `createCredentialOptions` to make the interface clearer and more extensible?

randmonkey (commented):
Generally LGTM, some minor comments.

kong/basic_auth_service.go
Prashansa-K: Ohh. I understood your point now. I will update the code.

PR #606, opened 2026-03-13
feat: Added CreateWithID func and test for graphql_rate_limiting_cost_decoration_service

shivaygupta-dotcom:
https://konghq.atlassian.net/browse/FTI-7307
https://github.com/Kong/deck/issues/1909

Summary:

1. Introduced a new method CreateWithID in AbstractGraphqlRateLimitingCostDecorationService to allow creating CostDecoration entities in Kong with a specified ID.
2. This change was required because the existing Create method does not allow specifying an ID, whereas decK supports user-defined IDs during resource creation.

### DIFF
--- a/kong/graphql_rate_limiting_cost_decoration_service_test.go
+++ b/kong/graphql_rate_limiting_cost_decoration_service_test.go
@@ -66,15 +66,18 @@ func TestGraphqlRateLimitingCostDecorationService(t *testing.T) {
 		createdDeco, err := client.GraphqlRateLimitingCostDecorations.CreateWithID(defaultCtx, deco)
 		require.NoError(t, err)
 		require.NotNil(t, createdDeco)
+		require.Equal(t, id, *createdDeco.ID)
 
 		deco, err = client.GraphqlRateLimitingCostDecorations.Get(defaultCtx, createdDeco.ID)
 		require.NoError(t, err)
 		require.NotNil(t, deco)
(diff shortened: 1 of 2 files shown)


kong/graphql_rate_limiting_cost_decoration_service_test.go
```diff
@@ -53,4 +53,43 @@ func TestGraphqlRateLimitingCostDecorationService(t *testing.T) {
...
+
+		deco, err = client.GraphqlRateLimitingCostDecorations.Get(defaultCtx, createdDeco.ID)
+		require.NoError(t, err)
+		require.NotNil(t, deco)
+
+		deco.TypePath = String("car.designation")
+		deco, err = client.GraphqlRateLimitingCostDecorations.Update(defaultCtx, deco)
+		require.NoError(t, err)
+		require.NotNil(t, deco)
+		require.Equal(t, "car.designation", *deco.TypePath)
```
Prashansa-K: Let's also check for ID equality since this function primarily deals with that.
We want to ensure that ID passed is same as the ID in the new entity created.

Let's check it after the GET call after create. Line numbers got messed up while adding this comment.

Prashansa-K:
> Added CRUD support for GraphQL Rate Limiting Advanced cost decorators in decK.

This line in summary doesn't relate with this particular change in this repo. We can remove it.

Prashansa-K (approved):

PR #618, opened 2026-05-31
fix: preserve "json"-type plugin config fields in fillConfigRecord

### LINKED ISSUE #524: Missing handling of plugin fields of type "json" in fillConfigRecord
https://github.com/Kong/go-kong/blob/2ea7f986af4a6c1f979b6ad5621d21cbd0a0fe9a/kong/utils.go#L419

`deck gateway sync` and `deck gateway diff` incorrectly print the result of the diff operation for plugins that declare any of their fields as type "json". This happens because the fillConfigRecord function does not handle those fields and, even when the field is set to a proper value, it updates it to null before performing the diff.

I have verified that a simple `if`  added to this function solves the issue:
```

		if ftype.String() == "json" {
			if _, ok := config[fname]; ok {
				return true
			}
		}
```

but i'm unsure if that is the best way to solve it hence no pull request

AdamBrauns:
## Summary

`fillConfigRecord` does not handle plugin config fields declared with the Kong `json` type. When such a field holds a value but has no schema default, it is overwritten with `nil` during default-filling. Because decK fills defaults when building the desired plugin state for a diff, this makes `deck gateway diff` / `deck gateway sync` report a **perpetual no-op `update`** for any plugin that uses a `json` field.

A concrete example is the `ai-mcp-proxy` plugin, whose `config.tools[].request_body` is a `json` field: decK nulls it in the desired state while the live config from Kong keeps it populated, so the two never compare equal.

Closes #524.

## Root cause

`fillConfigRecord` (`kong/utils.go`) only special-cases the `record`, `array`, and `set` field types. A populated `json` value (an arbitrary JSON object, i.e. a `map[string]interface{}`) falls through the "already set" detection — that branch only returns early for arbitrary `keys`/`values` maps — and reaches the default-assignment branch. Since a `json` field has no default, it is set to `nil`, discarding the user's value.

## Fix

In the `map[string]interface{}` arm of the "already set" switch, treat a populated `json` field as already fully set and return early. There is no sub-schema to recurse into for a `json` field, so the existing value should be preserved as-is. The change is scoped strictly to `type == "json"`, so it does not affect records, primitives, arrays, or `keys`/`values` maps.

## Testing

- Added `Test_FillPluginsDefaults_JSONType`, covering a populated `json` field preserved alongside normal default-filling of sibling fields, and distinct `json` values preserved across multiple array-of-record elements.
- Confirmed the new test **fails without the fix** (the `json` value comes back `nil`) and passes with it.
- The existing default-filling unit tests (`Test_fillConfigRecord`, `Test_FillPluginsDefaults*`, etc.) continue to pass — no regressions.
- Verified against a live Konnect dev deployment: with this fix in a locally built decK, the previously perpetual no-op `update` on the `ai-mcp-proxy` plugin no longer appears in `deck gateway diff`.


tao12345666333:
Thanks for your contributions! 
Your analysis in the PR description is very detailed, which is great. However, after investigation, I found that we can have a relatively general approach, so I opened PR #625 to replace this PR (sorry for not noticing this PR before opening that one).

This PR is also correct; however, through the modifications in #625, we can reuse the existing tests to make it more complete without needing to add additional test cases.

Next time, if we don't respond to your PR, it might be that we missed the notification. Please feel free to ping me.
Thank you again!

PR #619, opened 2026-06-01
feat: added support for custom and cloned plugin definitions

Prashansa-K:
Gateway 3.15 is introducing `Cloned Plugin Definitions` and `Custom Plugin Definitions`
as native entities that need to be supported by deck.
[PRD](https://docs.google.com/document/d/1jwuHoooUzpXkLv-zHSEj1HO-gA3N815gzAZNQVigAbM/edit?tab=t.0) is attached.

This PR introduces the two new entities, along with their tests.
Custom Plugin feature is gated behind `KONG_CUSTOM_PLUGINS_ENABLED: "on"`.
The kong container startup script is updated with this environment variable and
test-utilities are updated to check for this before going ahead with custom plugins test.

For: 
- https://github.com/Kong/deck/issues/2073 
- https://github.com/Kong/deck/issues/2074


kong/cloned_plugin_definition_service.go
```diff
@@ -0,0 +1,159 @@
...
+
+// Update updates a ClonedPluginDefintion in Kong.
+func (s *ClonedPluginService) Update(ctx context.Context,
+	plugin *ClonedPluginDefinition,
+) (*ClonedPluginDefinition, error) {
+	if isEmptyString(plugin.ID) {
+		return nil, fmt.Errorf("ID cannot be nil for Update operation")
+	}
+
+	endpoint := fmt.Sprintf("/cloned-plugins/%v", *plugin.ID)
```
shivaygupta-dotcom, line 89: Can we use `PUT` here instead of `PATCH`, as konnect doesnt support Patch, so when konnect gets onboarded here, we don't have any diff.

kong/custom_plugin_definition_service.go
```diff
@@ -0,0 +1,159 @@
...
+// Update updates a CustomPluginDefintion in Kong.
+func (s *CustomPluginService) Update(ctx context.Context,
+	plugin *CustomPluginDefinition,
+) (*CustomPluginDefinition, error) {
+	if isEmptyString(plugin.ID) {
+		return nil, fmt.Errorf("ID cannot be nil for Update operation")
+	}
+
+	endpoint := fmt.Sprintf("/custom-plugins/%v", *plugin.ID)
+	req, err := s.client.NewRequest("PATCH", endpoint, nil, plugin)
```
shivaygupta-dotcom: Can we use `PUT` here instead of `PATCH`, as konnect doesn't support Patch, so when konnect gets onboarded here, we don't have any diff.

Prashansa-K: I am clarifying this here: https://kongstrong.slack.com/archives/C04349E4KRC/p1780304229264699?thread_ts=1779349109.682309&cid=C04349E4KRC

shivaygupta-dotcom (approved):

PR #623, opened 2026-06-19
fix: do not fill auto fields with FillPluginsDefaults

pmalek:
Setting auto fields with KIC produced errors where, e.g. rate-limiting-advanced namespace field was set to `nil` which caused:

```
invalid config.namespace: required field missing
```

errors.

This change disables, filling auto fields in `FillPluginsDefaults` to prevent that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Table of Contents
 
+- [v0.76.2](#v0762)
+- [v0.76.1](#v0761)
 - [v0.76.0](#v0760)
 - [v0.75.1](#v0751)
 - [v0.75.0](#v0750)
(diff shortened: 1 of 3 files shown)


programmer04 (approved):

mheap:
This is a behaviour change for deck users. If we auto-generate a namespace their configuration is no longer deterministic.

Please make this an opt-in change

pmalek:
> This is a behaviour change for deck users. If we auto-generate a namespace their configuration is no longer deterministic.
> 
> Please make this an opt-in change

I will make an opt in change here.

In my understanding this is about auto generating when not provided only, which should be the case already (I may lack the understanding of our terraform provider though)

mheap:
Yes, that's correct. If the field is not provided we auto-generate. If an explicit null is provided we throw an error.

Both deck and TF send explicit null today

pmalek:
This is not needed, `FillPluginsDefaultsWithOpts` already allows this behavior through options.

PR #624, opened 2026-07-08
feat: support ai gateway models

harshadixit12:
In this PR, we are adding a new feature - using which
1. CRUD, List operations can be made on `AI model` entity on the AI Gateway 2.0
2. The same can be referenced on `Plugin` as a foreign key using key `model`

Along with this, we are adding tests, and a github workflow for AI Gateway tests.


kong/ai_model_service.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// Update updates an AI Model in Kong
+func (s *AIModelService) Update(ctx context.Context, aiModel *AIModel) (*AIModel, error) {
+	if isEmptyString(aiModel.ID) {
+		return nil, fmt.Errorf("ID cannot be nil for Update operation")
+	}
+
+	endpoint := fmt.Sprintf("/ai-models/%v", *aiModel.ID)
+	req, err := s.client.NewRequest("PATCH", endpoint, nil, aiModel)
```
Prashansa-K: Konnect doesn't have PATCH endpoints. So, we will have to check for that condition and use PUT as well.
Ref: https://github.com/Kong/go-kong/blob/main/kong/cloned_plugin_definition_service.go#L85

kong/plugin.go
```diff
@@ -20,6 +20,7 @@ type Plugin struct {
 	Protocols     []*string       `json:"protocols,omitempty" yaml:"protocols,omitempty"`
 	Tags          []*string       `json:"tags,omitempty" yaml:"tags,omitempty"`
 	Partials      []*PartialLink  `json:"partials,omitempty" yaml:"partials,omitempty"`
+	Model         *AIModel        `json:"model,omitempty" yaml:"model,omitempty"`
```
Prashansa-K, line 23: We should have tests for plugin+ai_model as well.

kong/ai_model_service_test.go
```diff
@@ -0,0 +1,178 @@
...
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAIModelsCRUD(T *testing.T) {
+	RunWhenDBMode(T, "postgres")
```
Prashansa-K: I believe we will add AI gateway in the test matrix and add an apt condition here, correct?

harshadixit12: yes - this is the only failing test i believe

kong/plugin.go
harshadixit12: Will add

.github/workflows/integration-test-ai-gateway.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - secret-available
+    if: needs.secret-available.outputs.ok
+    strategy:
+      matrix:
+        router_flavor:
+          - 'traditional_compatible'
+    env:
+      KONG_ROUTER_FLAVOR: ${{ matrix.router_flavor }}
+      KONG_IMAGE_REPO: kong/kong-ai-gateway-dev
+      KONG_IMAGE_TAG: '2.0.1-rc.1'
```
harshadixit12: cc: @Prashansa-K for visibility - the stable version `2.0.0` has some bugs, i'm following up on when the next stable version (non dev) will be out.
https://kongstrong.slack.com/archives/C0BEBVDL5TL/p1783919611238019

kong/test_utils.go
```diff
@@ -128,6 +130,69 @@ func SkipWhenEnterprise(t *testing.T) {
...
+	}
+
+	isKongAIGateway := isKongAIGateway(info["plugins"].(map[string]interface{}))
+	if !isKongAIGateway {
+		t.Skip("Not an AI Gateway instance, skipping")
+	}
+}
+
+func isKongAIGateway(pluginConfig map[string]interface{}) bool {
+	availablePlugins, ok := pluginConfig["available_on_server"].(map[string]interface{})
```
harshadixit12: There is no single config field we can rely on to differentiate between AI gateway and enterprise - so relying on existence of `ai-model-selector` plugin as a proxy - this is not available on API gateway.
https://kongstrong.slack.com/archives/C0BEBVDL5TL/p1783687758252459

harshadixit12:
@Prashansa-K there are a number of tests (~100) which are meant for API gateway, but are running for AI gateway because they are not gated with runwhenkong / runwhenenterprise - while they are passing, should we find a way to skip them in the AI gateway workflow?

.github/workflows/integration-test-ai-gateway.yaml
Prashansa-K: Should we keep this in a matrix rather than one fixed value, so as to allow easy expansion for future versions?

kong/ai_model_service.go
Prashansa-K: We aren't doing this?

kong/ai_model_service_test.go
```diff
@@ -0,0 +1,178 @@
...
+import (
+	"testing"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAIModelsCRUD(T *testing.T) {
+	RunWhenAIGateway(T, "<3.0.0")
```
Prashansa-K: Why <3.0.0 and not >2.0.0?

kong/test_utils.go
```diff
@@ -48,6 +48,8 @@ func RunWhenKong(t *testing.T, versionRange string) {
 	if !r(currentVersion) {
 		t.Skipf("kong version %s not in range %s", currentVersion, versionRange)
 	}
+
+	SkipWhenAIGateway(t)
```
Prashansa-K, line 53: Should we add this in RunWhenEnterprise func too?

kong/ai_model_service_test.go
harshadixit12: >=2.0.0  makes sense, let me update.

kong/test_utils.go
harshadixit12: yes we should, I'll update.

Prashansa-K:
> @Prashansa-K there are a number of tests (~100) which are meant for API gateway, but are running for AI gateway because they are not gated with runwhenkong / runwhenenterprise - while they are passing, should we find a way to skip them in the AI gateway workflow?

This should work when we add a skipAIGateway call in runWhenEnterprise, right?

kong/ai_model_service_test.go
Prashansa-K: IIRC, an = would make it rigid for exactly 2.0.0 and won't take minor or patch versions as well.

kong/ai_model_service.go
harshadixit12: Let me check, might have missed this comment.
We aren't using PATCH at all though - not even for update through GDR - even the other entities use create, but pass ID, leading to PUT.

kong/ai_model_service_test.go
harshadixit12: Markdown syntax 😅 
I meant `>=2.0.0`

kong/ai_model_service.go
Prashansa-K: Older entities tended to use that. All the newer ones we added, we used Update calls to avoid confusions.

harshadixit12:
> > @Prashansa-K there are a number of tests (~100) which are meant for API gateway, but are running for AI gateway because they are not gated with runwhenkong / runwhenenterprise - while they are passing, should we find a way to skip them in the AI gateway workflow?
> 
> This should work when we add a skipAIGateway call in runWhenEnterprise, right?

No - some of these tests are directly using GetVersionForTesting(T)

kong/ai_model_service.go
harshadixit12: Oh, I missed that, let me update in that case.

Prashansa-K:
> > > @Prashansa-K there are a number of tests (~100) which are meant for API gateway, but are running for AI gateway because they are not gated with runwhenkong / runwhenenterprise - while they are passing, should we find a way to skip them in the AI gateway workflow?
> > 
> > 
> > This should work when we add a skipAIGateway call in runWhenEnterprise, right?
> 
> No - some of these tests are directly using GetVersionForTesting(T)

It should be okay for now. Once we have a proper config key in the schema for AI gateways, we can add a if check. Let it be for now.

kong/ai_model_service.go
harshadixit12: @Prashansa-K for cloned plugins - `tags` can't be unset using deck (verified using v1.64) - this happens because of using PATCH with omitempty. I'll add the condition to use PUT with konnect, but continue to use the `create` for updates through deck.
Tracking the bug for cloned plugins[ here](https://github.com/Kong/deck/issues/2152).

kong/test_utils.go
```diff
@@ -128,6 +132,69 @@ func SkipWhenEnterprise(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	if !IsKongAIGateway {
+		t.Skip("Not an AI Gateway instance, skipping")
+	}
+}
+
+func IsKongAIGateway() (bool, error) {
```
harshadixit12, line 186: Reference for the approach: https://kongstrong.slack.com/archives/C0BEBVDL5TL/p1783928421178779?thread_ts=1783687758.252459&cid=C0BEBVDL5TL

.github/workflows/integration-test-ai-gateway.yaml
```diff
@@ -0,0 +1,90 @@
...
+  test-ai-gateway:
+    timeout-minutes: ${{ fromJSON(vars.GHA_DEFAULT_TIMEOUT || 10) }}
+    continue-on-error: true
+    needs:
+    - secret-available
+    if: needs.secret-available.outputs.ok
+    strategy:
+      matrix:
+        kong_image_tag:
+          - '29019036398-31e991be3666685cf27cfa3858438beaf7e52dae'
```
harshadixit12: Latest dev image: https://kongstrong.slack.com/archives/C0BEBVDL5TL/p1783928784139089

kong/test_utils.go
harshadixit12: It is public since being reused in GDR - I'm not super happy with this - but I do see other functions from here being used in GDR.

Prashansa-K (approved):

